PR #1, opened 2020-01-31
chore: add split repo templates

plamut:
This is "step 6 - Add Repo Templates" of the repo split procedure.

Not entirely sure if all `test_utils` files are needed, please check (removed just the egg-info metadata).


noxfile.py
```diff
@@ -113,9 +110,8 @@ def system(session):
     # Install all test dependencies, then install this package into the
     # virtualenv's dist-packages.
     session.install("mock", "pytest")
-    for local_dep in LOCAL_DEPS:
-        session.install("-e", local_dep)
-    session.install("-e", "../test_utils/")
+    # session.install("-e", "../test_utils/")
+    session.install("-e", "test_utils")
```
plamut: This was needed, because `test_utils` cannot be placed "one level up", as `synth.py` is already at the root level.

I intentionally left the old path just commented out (as apposed to removing it) to remind us that we need to update the noxfile template after the repo split is done.

PR #34, opened 2020-02-20
test: add system tests for common RBAC operations

### LINKED ISSUE #8: PubSub: Add more system tests covering various RBAC-related scenarios
A [regression](https://github.com/googleapis/google-cloud-python/issues/9339) introduced not so long ago could have been prevented if we had more RBAC-related system tests. A system test covering the fix was added in googleapis/google-cloud-python#9507.

Since PubSub backend defines several different [user roles](https://cloud.google.com/pubsub/docs/access-control#roles), system tests must be added to cover at least the most common use case scenarios that can be affected by the user role used in the session.

plamut:
Fixes #8.

This PR adds several system tests focusing on RBAC.

I used all non-admin [PubSub roles](https://cloud.google.com/pubsub/docs/access-control#roles) in common happy use cases. If there are other uses cases that should also be added, feel free to suggest in the comments.

### PR checklist:
- [ ] Appropriate docs were updated (if necessary) - discussed at a meeting, will be addressed separately

### DIFF
--- a/tests/system.py
+++ b/tests/system.py
@@ -628,6 +628,9 @@ class TestStreamingPull(object):
         response = subscriber.pull(subscription_path, max_messages=2)
         assert len(response.received_messages) == 2
 
+    @pytest.mark.skip(
+        "Snapshot creation is not instant on the backend, causing test falkiness."
+    )
     @pytest.mark.skipif(
         "KOKORO_GFILE_DIR" not in os.environ,
         reason="Requires Kokoro environment with a service account with limited role.",
--- a/tests/system.py
+++ b/tests/system.py
@@ -564,7 +564,7 @@ class TestStreamingPull(object):
 
     @pytest.mark.skipif(
         "KOKORO_GFILE_DIR" not in os.environ,
-        reason="Requires Kokoro environment with a limited subscriber service account.",
+        reason="Requires Kokoro environment with a service account with limited role.",
     )
     def test_streaming_pull_subscriber_permissions_sufficient(
         self, publisher, topic_path, subscriber, subscription_path, cleanup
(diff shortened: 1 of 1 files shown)


plamut:
Hmm, the snapshot seek test seems to be flaky. Are snapshot not "instant" by any chance?

**Update:** Yes, creating a snapshot is actually an async operation on the server, despite a "success" is returned. In the short term we need to disable such flaky test.

tests/system.py
```diff
@@ -601,6 +601,136 @@ def test_streaming_pull_subscriber_permissions_sufficient(
...
+    ):
+        snapshot_name = "snap" + unique_resource_id("-")
+        snapshot_path = "projects/{}/snapshots/{}".format(project, snapshot_name)
+
+        # Make sure the topic and subscription get deleted.
+        cleanup.append((publisher.delete_topic, topic_path))
+        cleanup.append((subscriber.delete_subscription, subscription_path))
+        cleanup.append((subscriber.delete_snapshot, snapshot_path))
+
+        # create a topic and subscribe to it
```
pradn: nite: capitalize

tests/system.py
```diff
@@ -601,6 +601,136 @@ def test_streaming_pull_subscriber_permissions_sufficient(
...
+        # A service account granting only the pubsub.viewer role must be used.
+        filename = os.path.join(
+            os.environ["KOKORO_GFILE_DIR"], "pubsub-viewer-service-account.json"
+        )
+        viewer_only_subscriber = type(subscriber).from_service_account_file(filename)
+        viewer_only_publisher = type(publisher).from_service_account_file(filename)
+
+        # The following operations should not raise permission denied errors.
+        # NOTE: At least one topic exists.
+        topic = next(iter(viewer_only_publisher.list_topics(project_path)))
```
pradn: Perhaps we should have negative tests too? "This account doesn't have this permission so we should expect an exception."

I don't know if it's worth the extra code. What do you think?

tests/system.py
```diff
@@ -601,6 +601,136 @@ def test_streaming_pull_subscriber_permissions_sufficient(
...
+        project_path = "projects/" + project
+
+        # Make sure the created topic gets deleted.
+        cleanup.append((publisher.delete_topic, topic_path))
+
+        publisher.create_topic(topic_path)
+
+        # A service account granting only the pubsub.viewer role must be used.
+        filename = os.path.join(
+            os.environ["KOKORO_GFILE_DIR"], "pubsub-viewer-service-account.json"
```
pradn: These tests are based on the IAM roles, which each encompass multiple permissions. It's okay for these tests to deal in roles and not the concrete permissions themselves since most of our customers don't create custom roles with exotic permission sets.

I just wanted to point it out so we're aware of the distinction between roles/permissions.

tests/system.py
plamut: I was thinking about it, but the number of test cases could explode, and I definitely didn't want to turn this into "unit testing with system tests".

The primary intent was to assure that a client _can_ perform common expected operations without requiring extra permissions - as was the case with that RBAC regression in 1.0.0 (IIRC) - and not that the backend correctly prevents the client from doing things that it's not supposed to. Negative tests are thus the backend's responsibility IMO.

tests/system.py
plamut: Yeah, didn't go into those specifics, as the number of test cases would explode. I also assumed most users will use predefined roles and not a custom mix of permissions.

tests/system.py
plamut:
```suggestion
        # Create a topic and subscribe to it.
```

plamut: Done. :)

pradn:
Good to merge!

PR #38, opened 2020-02-26
feat(pubsub): implement max_duration_per_lease_extension option

### LINKED ISSUE #35: pubsub: expose a max_duration_per_lease_extension setting for subscribers
For languages that support lease management, we have bound the percentile distribution to 10s - 10m. It would be useful if the upper bound were configurable.

A user might process messages very slowly (8m to ack)
If the application fails 5s after receiving a message (whose ack deadline was set to 8m, because that's the deadline the percentile distribution gave), the user has to wait 7m55s for the message to be redelivered
Users should be able to set `max_duration_per_lease_extension` = 20s, which caps an individual modAck RPC to 20s. This would have the downside of causing more modack RPCs to be sent, but the upside that the redelivery time would only be 15s
This setting differs from [`max_lease_duration`](https://github.com/googleapis/google-cloud-dotnet/blob/174e30c1ef39c5d1e1e51590b6af402aa0316c54/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs#L229), which defines the total amount of time we'll hold a message.

In addition, `max_duration_per_lease_extension` should be disabled if <=0, and should default to 0.

IlyaFaer:
Closes #35

### DIFF
--- a/google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
+++ b/google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
@@ -130,8 +130,8 @@ class Leaser(object):
             # Determine the appropriate duration for the lease. This is
             # based off of how long previous messages have taken to ack, with
             # a sensible default and within the ranges allowed by Pub/Sub.
-            p99 = self._manager.ack_histogram.percentile(99)
-            _LOGGER.debug("The current p99 value is %d seconds.", p99)
+            deadline = self._manager.ack_deadline
+            _LOGGER.debug("The current deadline value is %d seconds.", deadline)
 
             # Make a copy of the leased messages. This is needed because it's
             # possible for another thread to modify the dictionary while
@@ -173,7 +173,7 @@ class Leaser(object):
                 #       way for ``send_request`` to fail when the consumer
                 #       is inactive.
                 self._manager.dispatcher.modify_ack_deadline(
-                    [requests.ModAckRequest(ack_id, p99) for ack_id in ack_ids]
+                    [requests.ModAckRequest(ack_id, deadline) for ack_id in ack_ids]
                 )
 
             # Now wait an appropriate period of time and do this again.
@@ -182,7 +182,7 @@ class Leaser(object):
             # period between 0 seconds and 90% of the lease. This use of
             # jitter (http://bit.ly/2s2ekL7) helps decrease contention in cases
             # where there are many clients.
-            snooze = random.uniform(0.0, p99 * 0.9)
+            snooze = random.uniform(0.0, deadline * 0.9)
             _LOGGER.debug("Snoozing lease management for %f seconds.", snooze)
             self._stop_event.wait(timeout=snooze)
 
--- a/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
+++ b/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
@@ -198,13 +198,12 @@ class StreamingPullManager(object):
 
         if hist_size > target_size:
             self._last_histogram_size = hist_size
-            p99 = self.ack_histogram.percentile(percent=99)
+            self._ack_deadline = self.ack_histogram.percentile(percent=99)
 
-            if self.flow_control.max_extension_period > 0:
-                self._ack_deadline = min(p99, self.flow_control.max_extension_period)
-            else:
-                self._ack_deadline = p99
-
+        if self.flow_control.max_extension_period > 0:
+            self._ack_deadline = min(
+                self._ack_deadline, self.flow_control.max_extension_period
+            )
         return self._ack_deadline
 
     @property
--- a/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
+++ b/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
@@ -200,9 +200,9 @@ class StreamingPullManager(object):
             self._last_histogram_size = hist_size
             self._ack_deadline = self.ack_histogram.percentile(percent=99)
 
-        if self.flow_control.max_extension_period > 0:
+        if self.flow_control.max_duration_per_lease_extension > 0:
             self._ack_deadline = min(
-                self._ack_deadline, self.flow_control.max_extension_period
+                self._ack_deadline, self.flow_control.max_duration_per_lease_extension
             )
         return self._ack_deadline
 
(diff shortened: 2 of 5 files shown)


tests/unit/pubsub_v1/subscriber/test_leaser.py
```diff
@@ -84,6 +84,7 @@ def create_manager(flow_control=types.FlowControl()):
     manager.is_active = True
     manager.flow_control = flow_control
     manager.ack_histogram = histogram.Histogram()
+    manager.ack_deadline = 10
```
IlyaFaer: Add `ack_deadline` into Manager mock. Otherwise few tests will fail because of comparison `int` > `mock.Mock()`

tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
```diff
@@ -144,6 +144,15 @@ def test_ack_deadline():
     assert manager.ack_deadline == 20
 
 
+def test_ack_deadline_w_max_extension_period():
+    manager = make_manager()
+    manager._flow_control = types.FlowControl(max_extension_period=5)
+    for _ in range(5):
+        manager.ack_histogram.add(20)
+
+    assert manager.ack_deadline == 5
```
IlyaFaer: Adding `20` five times in a row to fill histogram with data. With this, `max_extension_period` should be chosen

google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
```diff
@@ -130,8 +130,8 @@ def maintain_leases(self):
             # Determine the appropriate duration for the lease. This is
             # based off of how long previous messages have taken to ack, with
             # a sensible default and within the ranges allowed by Pub/Sub.
-            p99 = self._manager.ack_histogram.percentile(99)
-            _LOGGER.debug("The current p99 value is %d seconds.", p99)
+            deadline = self._manager.ack_deadline
+            _LOGGER.debug("The current deadline value is %d seconds.", deadline)
```
IlyaFaer: Getting deadline from the manager's method

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -191,9 +191,20 @@ def ack_deadline(self):
...
+        hist_size = len(self.ack_histogram)
+
+        if hist_size > target_size:
+            self._last_histogram_size = hist_size
+            p99 = self.ack_histogram.percentile(percent=99)
+
+            if self.flow_control.max_extension_period > 0:
+                self._ack_deadline = min(p99, self.flow_control.max_extension_period)
+            else:
+                self._ack_deadline = p99
```
IlyaFaer: As I already mentioned in original issue, I see an error here: `_last_histogram_size` is never changed. According to the method docs, it should be implemented like this. We're remembering histogram size only if we got enough data. Otherwise we can get in a situation when data added with very small portions a lot of times - in total we got a lot of data, but we're not processing it, 'cause growth steps were small

pradn: Yes, that was a bug. Good catch!

tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
pradn: Can you also add 
"assert manager.ack_deadline == 5"
before line 150?

This way we check that the max_extension_period is respected even if there have been no add() calls to ack_histogram.

IlyaFaer: Oh, if `max_extension_period` should be respected from the very beginning, then property can be written little bit easier. Pushed

google/cloud/pubsub_v1/types.py
```diff
@@ -112,6 +114,10 @@
         "The maximum amount of time in seconds to hold a lease on a message "
         "before dropping it from the lease management."
     )
+    FlowControl.max_extension_period.__doc__ = (
+        "Bounds the maximum amount of time in seconds before a message redelivery "
```
pradn: Maybe start with:
"The max amount of time for a single lease extension/renewal attempt."

pradn:
It looks pretty good, just a small comment.

Let's wait on merging this in until we have word on the final name for "max_extension_period".

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -191,9 +191,19 @@ def ack_deadline(self):
...
+        )
+        hist_size = len(self.ack_histogram)
+
+        if hist_size > target_size:
+            self._last_histogram_size = hist_size
             self._ack_deadline = self.ack_histogram.percentile(percent=99)
+
+        if self.flow_control.max_duration_per_lease_extension > 0:
+            self._ack_deadline = min(
+                self._ack_deadline, self.flow_control.max_duration_per_lease_extension
```
pradn: nit, can you split line 205 into two lines, please?

google/cloud/pubsub_v1/types.py
```diff
@@ -112,6 +119,11 @@
         "The maximum amount of time in seconds to hold a lease on a message "
         "before dropping it from the lease management."
     )
+    FlowControl.max_duration_per_lease_extension.__doc__ = (
+        "The max amount of time in seconds for a single lease extension attempt. "
+        "Bounds the delay before a message redelivery in the event the "
```
pradn: super nit: replace "in the event" with "if"

tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
```diff
@@ -144,6 +144,17 @@ def test_ack_deadline():
     assert manager.ack_deadline == 20
 
 
+def test_ack_deadline_w_max_duration_per_lease_extension():
```
pradn: nit: i prefer "w" being "with"

google/cloud/pubsub_v1/types.py
IlyaFaer: Done

tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
IlyaFaer: Done

PR #40, opened 2020-02-28
fix: shutdown error on streaming pull callback error

### LINKED ISSUE #39: RunTime error occurs when shutting down the streaming pull manager on user callback exception
@plamut i no longer receive the deadline exceeded error #36  however i dont see the actions being perfomed under 
def callback(message)
Can you please help on this

plamut:
Fixes #39.

This PR fixes an error that can happen if cancelling a streaming pull future when the streaming pull manager has already been shut down due to the error in the user callback.

### How to test

- Have a topic and a subscription, publish at least one message to the topic.
- Take the basic [streaming pull sample](https://cloud.google.com/pubsub/docs/pull#asynchronous-pull) from the docs.
- Adjust the message received callback to raise an exception

**Actual result (before the fix):**
An error occurs during the shutdown:
```
RuntimeError: set_result can only be called once.
```

**Expected result (after the fix):**
No internal error occurs during the shutdown, except for the error from the user callback, as expected.

### PR checklist

### DIFF
--- a/tests/unit/pubsub_v1/subscriber/test_futures_subscriber.py
+++ b/tests/unit/pubsub_v1/subscriber/test_futures_subscriber.py
@@ -57,6 +57,18 @@ class TestStreamingPullFuture(object):
 
         assert not future.running()
 
+    def test__on_close_callback_future_already_done(self):
+        future = self.make_future()
+
+        future.set_result("foo")
+        assert future.done()
+
+        # invoking on close callback should result in an error
+        future._on_close_callback(mock.sentinel.manager, "bar")
+
+        result = future.result()
+        assert result == "foo"  # on close callback was a no-op
+
     def test_cancel(self):
         future = self.make_future()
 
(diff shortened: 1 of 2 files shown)


tests/unit/pubsub_v1/subscriber/test_futures_subscriber.py
```diff
@@ -57,6 +57,18 @@ def test__on_close_callback_failure(self):
 
         assert not future.running()
 
+    def test__on_close_callback_future_already_done(self):
+        future = self.make_future()
+
+        future.set_result("foo")
+        assert future.done()
+
+        # invoking on close callback should result in an error
```
pradn: "should not"?

plamut: :self-facepalm: :)

plamut:
```suggestion
        # invoking on close callback should not result in an error
```

PR #49, opened 2020-03-20
fix: don't assert on unordered publishes after publish error

### LINKED ISSUE #48: asserting throwing for async publish with currently errored batch
#### Environment details

  - OS type and version: linux
  - Python version: `python --version` : Python 3.7.6
  - pip version: `pip --version` :  pip 19.3.1
  - `google-cloud-pubsub` version: `pip show google-cloud-pubsub` 1.3.1

#### Code example

```python
import time


def main():
    topic_path = "projects/*****/topics/*****"

    try:
        publisher.publish(topic_path, data=b'')  # invalid
    except:
        pass

    while True:
        print('publishing')
        publisher.publish(topic_path, data=b'valid')
        time.sleep(1)


if __name__ == '__main__':
    main()
```

#### Stack trace
```
  File "pubsub_error.py", line 21, in <module>
    main()
  File "pubsub_error.py", line 16, in main
    publisher.publish(topic_path, data=b'valid')
  File "/Users/camerondavison/Library/Python/3.7/lib/python/site-packages/google/cloud/pubsub_v1/publisher/client.py", line 364, in publish
    future = sequencer.publish(message)
  File "/Users/camerondavison/Library/Python/3.7/lib/python/site-packages/google/cloud/pubsub_v1/publisher/_sequencer/unordered_sequencer.py", line 113, in publish
    future = batch.publish(message)
  File "/Users/camerondavison/Library/Python/3.7/lib/python/site-packages/google/cloud/pubsub_v1/publisher/_batch/thread.py", line 334, in publish
    ), "Publish after stop() or publish error."
AssertionError: Publish after stop() or publish error.
```

As I mentioned just now in https://github.com/googleapis/python-pubsub/pull/26#issuecomment-601333328 it looks like there is an assert that was added that can now trigger for an errored out batch.

batch errors could happen any time so it seems like it would be better to move the assert below the check to see if the batch is still accepting new messages so that a new batch can be created if it errored out.

I validated the same code does not throw an exception on publish in `1.2.0`

pradn:
Unordered batches that threw errors on invoking the publish RPC led to subsequent assertion failures in thread.py::publish(). These assertion failures were legitimate. Publish calls should not be called on batches that have already been committed. (In this case, they were committed and then returned an error.)

This pull request fixes the underlying issue by clearing out the UnorderedSequencer's _current_batch on commit(). This ensures that no subsequent publish() calls will be called on that batch since the reference to it is lost.

I add a new "test_publish_after_batch_error" test for this scenario.

The OrderedSequencer does not have this issue. On error, it cancels all batches and pauses the ordering key. The existing "test_batch_done_unsuccessfully" test captures this scenario.

Fixes #48

### DIFF
--- a/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -369,3 +369,6 @@ class Batch(base.Batch):
             self.commit()
 
         return future
+
+    def _set_status(self, status):
+        self._status = status
--- a/google/cloud/pubsub_v1/publisher/_sequencer/unordered_sequencer.py
+++ b/google/cloud/pubsub_v1/publisher/_sequencer/unordered_sequencer.py
@@ -65,6 +65,12 @@ class UnorderedSequencer(base.Sequencer):
         if self._current_batch:
             self._current_batch.commit()
 
+            # At this point, we lose track of the old batch, but we don't
+            # care since we just committed it.
+            # Setting this to None guarantees the next publish() creates a new
+            # batch.
+            self._current_batch = None
+
     def unpause(self):
         """ Not relevant for this class. """
         raise NotImplementedError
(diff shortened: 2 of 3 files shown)


plamut:
(reviewing right now, will merge myself if all is fine and CI is done by then)

pradn:
> (reviewing right now, will merge myself if all is fine and the CI is done by then)

Thanks Peter!

PR #51, opened 2020-03-23
test: put RBAC tests into their own suite

### LINKED ISSUE #50: chore: group RBAC system tests into their own suite
Currently all RBAC system tests are placed under the `TestStreamingPull` class, even if they don't use streaming pull at all. These tests should be moved to their own class.

plamut:
Closes #50.

This PR reorganizes a few system tests, noticed it while adding a new (unrelated) system test.

### PR checklist


tests/system.py
```diff
@@ -731,30 +717,29 @@ def test_editor_role_can_create_resources(
...
 
-        for batch_size in batch_sizes:
-            msg_batch = self._make_messages(count=batch_size)
-            for msg in msg_batch:
-                future = publisher.publish(
-                    topic_path, msg, seq_num=str(next(msg_counter))
-                )
-                publish_futures.append(future)
-            time.sleep(0.1)
+def _publish_messages(publisher, topic_path, batch_sizes):
```
pradn: nit: would be nice to have helper functions at the top, instead of at the bottom

plamut: I actually prefer the opposite, i.e. having the higher-level code at the top, and the callees topologically sorted towards the bottom, makes it easier to  follow the code in a drill-down fashion (IMO at least). One can then quickly check the high-level tests, without having to first scroll through their helpers ("implementation details"), thus I would keep the existing structure of the file.

pradn: Okay, that's fine.

PR #53, opened 2020-03-23
test: cover the fix for sync pull with no messages

### LINKED ISSUE #52: Add test for 504 DeadlineExceeded for synchronous pull without published messages
Related to #47 and the [pull request](https://github.com/googleapis/python-pubsub/pull/43) that will fix it.

We should cover the fix with a system test to detect if a regression occurs again.

plamut:
Closes #52.

This PR is a follow-up to #43, requires the latter to be merged first.

To test, simply verify that the test fails before the fix in linked PR, and passes after merging it.

### PR checklist


tests/system.py
```diff
@@ -432,6 +433,28 @@ def one_arg_close(subscriber):  # the cleanup helper expects exactly one argumen
...
+
+    # Create a topic and subscribe to it.
+    publisher.create_topic(topic_path)
+    subscriber.create_subscription(subscription_path, topic_path)
+
+    try:
+        response = subscriber.pull(subscription_path, max_messages=2)
+    except core_exceptions.DeadlineExceeded:
+        pytest.fail(
+            "Unexpected DeadlineExceeded error on syncronous pull when no "
```
pradn: nit: spelling "synchronous"

plamut: Thanks, fixed.

PR #55, opened 2020-03-23
chore: add psutil to synth.py

hongalex:


plamut:
Process question - should we also regenerate the files every time when `synth.py` is updated, and commit those changes, too?

busunkim96:
@plamut Generally yes, but I think in this case the dependency was already in noxfile.py.

plamut:
That's true, its not strictly necessary to manually regenerate now, can wait until the next autosynth.

PR #58, opened 2020-03-24
feat: send client id with StreamingPullRequest

### LINKED ISSUE #62: Send client id with StreamingPullRequest
- The client id should be created randomly for each StreamingPullManager and will be used to establish affinity across stream disconnections/retries.
- Server-client affinity is important for ordering keys, where the backend tries to send the same keys to the same client.

pradn:
- The client id is created randomly for each StreamingPullManager and is used to establish affinity across stream disconnections/retries.
- Server-client affinity is important for ordering keys, where the backend tries to send the same keys to the same client.

Fixes #62


plamut:
What about a test? :)
Looks good otherwise.

pradn:
Okay, added tests!

plamut:
Oh, just a sanity check - is it indeed enough to send client ID just in the initial `StreamingPullRequest`? If not, we should include it in several other StreamingPullRequests the streming pull manager sends.

pradn:
Yes, sending client_id is only required on the first StreamingPullRequest. See the [field description](https://github.com/googleapis/python-pubsub/blob/master/google/cloud/pubsub_v1/proto/pubsub.proto#L1025) in the proto file.

pradn:
WIll wait for Kamal's review before merging.

plamut:
@pradn Ping, just checking the status of this.

kamalaboulhosn:
Sorry, I had lost track of this. Thanks for adding it!

PR #59, opened 2020-03-25
fix: use client_options.api_endpoint parameter instead of ignoring it

### LINKED ISSUE #61: The client_options.api_endpoint parameter is being ignored in both PublisherClient and SubscriberClient
This bug prevents users from using regional endpoints. It also makes it harder to test code against nonprod endpoints.

pradn:
- The client_options.api-endpoint parameter was being ignored. This bug prevented users from using regional endpoints. It also made it harder to test code against nonprod endpoints.

Fixes #61

### DIFF
--- a/google/cloud/pubsub_v1/publisher/client.py
+++ b/google/cloud/pubsub_v1/publisher/client.py
@@ -133,6 +133,12 @@ class Client(object):
                 target=os.environ.get("PUBSUB_EMULATOR_HOST")
             )
 
+        client_options = kwargs.pop("client_options", None)
+        if client_options and type(client_options["api_endpoint"]) is str:
+            self._target = client_options["api_endpoint"]
+        else:
+            self._target = publisher_client.PublisherClient.SERVICE_ADDRESS
+
         # Use a custom channel.
         # We need this in order to set appropriate default message size and
         # keepalive options.
@@ -217,7 +223,7 @@ class Client(object):
         Returns:
             str: The location of the API.
         """
-        return publisher_client.PublisherClient.SERVICE_ADDRESS
+        return self._target
 
     def _get_or_create_sequencer(self, topic, ordering_key):
         """ Get an existing sequencer or create a new one given the (topic,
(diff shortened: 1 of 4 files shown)


google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -133,6 +133,12 @@ def __init__(self, batch_settings=(), publisher_options=(), **kwargs):
                 target=os.environ.get("PUBSUB_EMULATOR_HOST")
             )
 
+        client_options = kwargs.pop("client_options", None)
+        if client_options and type(client_options["api_endpoint"]) is str:
```
plamut: Two things:
- If `api_endpoint` is a unicode string under Python 2, the type check will be negative and the option will (still) be ignored:
    ```py
    >>> type(u"foo/bar.com") is str
    False
    ```
    It is thus better to rewrite it as `isinstance(maybe_string, six.string_types)` (it does return `False` for `bytes` under Python 3, but this is fine, as the latter makes a clear distinction between text and binary data).
- if `client_options` does not contain the `"api_endpoint"` key, a `KeyError` will be raised, which is probably undesired, and we should just fall back to the default SERVICE_ADDRESS.

Let's thus rewrite this snippet, and add a test or two for the unhappy path, too. Applies to the subscriber client below, too.

pradn: Thanks, done!

tests/unit/pubsub_v1/publisher/test_publisher_client.py
```diff
@@ -63,6 +63,25 @@ def test_init_w_api_endpoint():
     ) == "testendpoint.google.com"
 
 
+def test_init_w_unicode_api_endpoint():
+    client_options = {"api_endpoint": u"testendpoint.google.com"}
```
plamut: We should also check the non-unicode string value (the one that can be problematic in Python 2), i.e. `{"api_endpoint": "testendpoint.google.com"}`.

pradn: I do that in the test above, no?

plamut: Ah, of course, late hour. It's the unicode version that would fail the `type(...) is str` check, nevermind.

PR #92, opened 2020-04-29
chore: remove local test_utils directory

### LINKED ISSUE #91: Remove test_utils directory
Now that test utils are available (and installed from) a GitHub repo, the `test_utils` directory is no longer needed and can be removed.

plamut:
Closes #91 🦕

### PR checklist


pradn:
Can you explain what these are needed for? How does one install them from a git repo?

plamut:
@pradn In the PubSub lib system tests use the `unique_resource_id` helper from it (and some other libraries rely on them even more). These utils were available as a separate directory in the old monorepo, but that broke after the repo split. Having a local copy in each of the new repository was a temporary workaround.

Test utils are now available in the common standalone repo, and I noticed that the noxfile already [uses that](https://github.com/googleapis/python-pubsub/blob/master/noxfile.py#L114), meaning that the local copy is not needed anymore.

PR #94, opened 2020-04-30
chore: install google-cloud-testutils by package name

busunkim96:
VPCSC is currently unable to run these tests because testutils is being installed through VCS.


busunkim96:
I didn't notice #92 🙃

PR #96, opened 2020-05-07
feat: add flow control for message publishing

### LINKED ISSUE #16: Pubsub: Consider a blocking publisher to prevent OOMing
Publishing medium-sized (32 kiB) messages in a loop causes the process's memory usage to quickly run out. I set a 2 GiB memory limit, and the process runs out of memory after publishing about 20000 messages. I believe there are two issues:

1. There is no limit to the number of queued bytes in the process.
2. The client library keeps 3 copies of the messages and retains them until the batch "monitor" thread finishes.

As far as I can see, there no good way to use the current client library to publish large batches of messages without HUGE memory consumption. I believe this can be fixed in two ways:

#### Reduce memory of the current library:

1. Release messages when the batch is done: In `thread.Batch._commit`, set `self._messages` to `None` after it is published. This *seems* to make memory consumption better, but I didn't test it super carefully.

2. Eliminate an extra copy: Change thread.Batch to build a PubsubRequest, and pass *that* into the gapic.PublisherClient, rather than a list of messages that then need to be copied. This reduces one copy of messages, which seems to make memory consumption better (and possibly make things faster? again not tested carefully)

#### Back Pressure

Change the publisher to keep a queue of batches, and block after there are more than N in flight at once.

I implemented a client that takes the blocking queue approach. It can publish this workload without ever exceeding 300 MiB of memory. Only 3 in-flight batches were necessary to have the process be CPU bound when running in Google Cloud, so I don't think the queue doesn't even need to be too large.

#### Environment details

*OS*: Linux, ContainerOS (GKE), Container is Debian9 (using distroless)
*Python*: 3.5.3
*API*: google-cloud-python 0.41.0

#### Steps to reproduce

1. Package the following program in a Docker container.
2. Run it in Kubernetes with a memory limit of 2 GiB
3. Watch it blow up when you run it.

#### Code example

```python
#!/usr/bin/env python3
'''
Create the topic and a subscription:

gcloud pubsub topics create deleteme_publish_memleak_test
gcloud pubsub subscriptions create deleteme_memleak_subscription --topic=deleteme_publish_memleak_test
'''

import argparse
import google.auth
import logging
import time


def main():
    parser = argparse.ArgumentParser(description='mass publish to pub/sub')
    parser.add_argument('--topic',
                        default='deleteme_publish_memleak_test',
                        help='topic to publish to')
    parser.add_argument('--messages',
                        type=int,
                        default=200000,
                        help='number of messages to publish')
    parser.add_argument('--bytes_per_message',
                        type=int,
                        default=32 * 1024,
                        help='bytes per message to publish')
    parser.add_argument('--log_every_n', type=int, default=2000, help='log every N messages')
    args = parser.parse_args()

    credentials, project_id = google.auth.default()
    publish_client = pubsub_v1.PublisherClient(credentials=credentials)
    topic_path = publish_client.topic_path(project_id, args.topic)

    logging.basicConfig(
        level=logging.DEBUG,
        format='%(asctime)s %(levelname)s %(name)s: %(message)s',
    )
    logging.info('publishing %d messages to topic %s; %d bytes per message; %.1f total MiB',
                 args.messages, topic_path, args.bytes_per_message,
                 (args.messages * args.bytes_per_message) / 1024. / 1024.)

    start = time.time()
    for i in range(args.messages):
        message = (i % 256).to_bytes(1, byteorder='little') * args.bytes_per_message
        publish_client.publish(topic_path, message)
        if i % args.log_every_n == 0:
            logging.info('published %d messages', i)

    end = time.time()
    logging.info('done publishing in %f s', end - start)


if __name__ == '__main__':
    main()
```

### LINKED ISSUE #72: PubSub Publisher 504 Deadline Exceeded when sending bulk messages
#### Environment details

  - OS type and version: `PRETTY_NAME="Debian GNU/Linux 10 (buster)"`
  - Python version: `python --version` python3.7
  - pip version: `pip --version`  20.0.2
  - `google-cloud-pubsub` version: `pip show google-cloud-pubsub` 1.4.2

#### Steps to reproduce

  1. Send large number of messages to a PubSub Topic fails when I try to send 5000 without pausing and waiting, for example.
  2. Check logs

#### Code example

```python
# I have also tried this using the callback and hit the same error. This is what I'm doing now though.
iter_size = 5000
iter_count = 0
futures = []
for row in rows:
     data = json.dumps(row, default=lambda x: str(x)).encode('utf-8')
     future = publish_client.publish(topic, data, origin='xxx')
     futures.append(future)
     published_count += 1
     iter_count += 1
     if iter_count >= iter_size:
           [f.result() for f in futures]
           futures = []
           iter_count = 0
     count += 1
     [f.result() for f in futures]
```

#### Stack trace
```
[2020-04-14 14:19:23,788] {{thread.py:264}} ERROR - Failed to publish 141 messages.
File "/usr/local/lib/python3.7/site-packages/google/api_core/grpc_helpers.py", line 57, in error_remapped_callable
     return callable_(*args, **kwargs)
   File "/usr/local/lib/python3.7/site-packages/grpc/_channel.py", line 826, in __call__
     return _end_unary_response_blocking(state, call, False, None)
   File "/usr/local/lib/python3.7/site-packages/grpc/_channel.py", line 729, in _end_unary_response_blocking
     raise _InactiveRpcError(state)
 grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
 	status = StatusCode.DEADLINE_EXCEEDED
 	details = "Deadline Exceeded"
 	debug_error_string = "{"created":"@1586873963.056617563","description":"Error received from peer ipv4:108.177.122.95:443","file":"src/core/lib/surface/call.cc","file_line":1056,"grpc_message":"Deadline Exceeded","grpc_status":4}"
 >
```

Other Details:

I've found a similar issue here 
https://github.com/googleapis/google-cloud-java/issues/3867

Which references to this issue
https://github.com/googleapis/google-cloud-java/issues/3003

Then
https://github.com/googleapis/java-pubsub/issues/31

This will normally be running on GKE, so I understand me running it locally may have some variation of behavior due to network speed. However, we would prefer to avoid having this issue either way.

Is this something that is planned/intended to be supported or should I implement auto-throttling?

plamut:
Closes #16.
Closes #72.

This PR adds flow control settings to publisher client, similar to what was [added](https://github.com/googleapis/java-pubsub/pull/119) to the Java client recently.

The PR does not interfere with existing publish logic (sequencers, ordering keys, batching...). Instead, it acts like a valve placed in front of the publish pipeline, and taking the configured action if it detects excessive message flow.

### Things left to do / discuss
- [ ] A system test for this? Although not sure how feasible would it be, as we would have to publish _massive_ amounts of data on fast connection to have a chance of triggering the flow control error...
- [ ] If the desired behavior is BLOCK, should blocking also support timeouts? However, every publish future gets resolved sooner or later be done (even if due to an error/timeout), and will release the corresponding message from the flow controller then.
 - [ ] Unlike Java, the `FlowControl` settings are placed under `PublisherOptions`, and not `BatchSettings`. They fit there much more naturally IMO, but I can change this.
- [ ] What should be the default `FlowControl` values (the thresholds, specifically)? I set the message count and byte size to 10x the defaults in BatchSettings, would that be fine for most users?
 - [ ] Documenting the new feature more extensively than docstring updates in this PR? Can in principle also be done in a separate PR.

### PR checklist

### DIFF
--- a/google/cloud/pubsub_v1/publisher/client.py
+++ b/google/cloud/pubsub_v1/publisher/client.py
@@ -343,11 +343,6 @@ class Client(object):
 
             pubsub_v1.publisher.exceptions.MessageTooLargeError: If publishing
                 the ``message`` would exceed the max size limit on the backend.
-
-            :exception:`~pubsub_v1.publisher.exceptions.FlowControlLimitError`:
-                If publishing a new message would exceed the publish flow control
-                limits and the desired action on overflow is
-                :attr:`~google.cloud.pubsub_v1.types.LimitExceededBehavior.ERROR`.
         """
         # Sanity check: Is the data being sent as a bytestring?
         # If it is literally anything else, complain loudly about it.
--- a/google/cloud/pubsub_v1/publisher/client.py
+++ b/google/cloud/pubsub_v1/publisher/client.py
@@ -34,6 +34,7 @@ from google.cloud.pubsub_v1.gapic.transports import publisher_grpc_transport
 from google.cloud.pubsub_v1.publisher._batch import thread
 from google.cloud.pubsub_v1.publisher._sequencer import ordered_sequencer
 from google.cloud.pubsub_v1.publisher._sequencer import unordered_sequencer
+from google.cloud.pubsub_v1.publisher.flow_controller import FlowController
 
 __version__ = pkg_resources.get_distribution("google-cloud-pubsub").version
 
@@ -93,7 +94,11 @@ class Client(object):
 
             # Optional
             publisher_options = pubsub_v1.types.PublisherOptions(
-                enable_message_ordering=False
+                enable_message_ordering=False,
+                flow_control=pubsub_v1.types.PublishFlowControl(
+                    message_limit=2000,
+                    limit_exceeded_behavior=pubsub_v1.types.LimitExceededBehavior.BLOCK,
+                ),
             ),
 
             # Optional
@@ -197,6 +202,9 @@ class Client(object):
         self._is_stopped = False
         # Thread created to commit all sequencers after a timeout.
         self._commit_thread = None
+
+        # The object controlling the message publishing flow
+        self._flow_controller = FlowController(self.publisher_options.flow_control)
 
     @classmethod
     def from_service_account_file(cls, filename, batch_settings=(), **kwargs):
@@ -333,6 +341,11 @@ class Client(object):
 
             pubsub_v1.publisher.exceptions.MessageTooLargeError: If publishing
                 the ``message`` would exceed the max size limit on the backend.
+
+            :exception:`~pubsub_v1.publisher.exceptions.FlowControlLimitError`:
+                If publishing a new message would exceed the publish flow control
+                limits and the desired action on overflow is
+                :attr:`~google.cloud.pubsub_v1.types.LimitExceededBehavior.ERROR`.
         """
         # Sanity check: Is the data being sent as a bytestring?
         # If it is literally anything else, complain loudly about it.
@@ -364,6 +377,13 @@ class Client(object):
             data=data, ordering_key=ordering_key, attributes=attrs
         )
 
+        # Messages should go through flow control to prevent excessive
+        # queuing on the client side (depending on the settings).
+        self._flow_controller.add(message)
+
+        def on_publish_done(future):
+            self._flow_controller.release(message)
+
         with self._batch_lock:
             if self._is_stopped:
                 raise RuntimeError("Cannot publish on a stopped publisher.")
@@ -372,6 +392,7 @@ class Client(object):
 
             # Delegate the publishing to the sequencer.
             future = sequencer.publish(message)
+            future.add_done_callback(on_publish_done)
 
             # Create a timer thread if necessary to enforce the batching
             # timeout.
--- a/google/cloud/pubsub_v1/publisher/client.py
+++ b/google/cloud/pubsub_v1/publisher/client.py
@@ -31,6 +31,8 @@ from google.cloud.pubsub_v1 import _gapic
 from google.cloud.pubsub_v1 import types
 from google.cloud.pubsub_v1.gapic import publisher_client
 from google.cloud.pubsub_v1.gapic.transports import publisher_grpc_transport
+from google.cloud.pubsub_v1.publisher import exceptions
+from google.cloud.pubsub_v1.publisher import futures
 from google.cloud.pubsub_v1.publisher._batch import thread
 from google.cloud.pubsub_v1.publisher._sequencer import ordered_sequencer
 from google.cloud.pubsub_v1.publisher._sequencer import unordered_sequencer
@@ -379,7 +381,12 @@ class Client(object):
 
         # Messages should go through flow control to prevent excessive
         # queuing on the client side (depending on the settings).
-        self._flow_controller.add(message)
+        try:
+            self._flow_controller.add(message)
+        except exceptions.FlowControlLimitError as exc:
+            future = futures.Future()
+            future.set_exception(exc)
+            return future
 
         def on_publish_done(future):
             self._flow_controller.release(message)
--- a/google/cloud/pubsub_v1/publisher/flow_controller.py
+++ b/google/cloud/pubsub_v1/publisher/flow_controller.py
@@ -100,18 +100,14 @@ class FlowController(object):
                 # Raising an error means rejecting a message, thus we do not
                 # add anything to the existing load, but we do report the would-be
                 # load if we accepted the message.
-                msg = (
-                    "Flow control limits would be exceeded "
-                    "(messages: {} / {}, bytes: {} / {})."
-                ).format(
-                    self._message_count + 1,
-                    self._settings.message_limit,
-                    self._total_bytes + message.ByteSize(),
-                    self._settings.byte_limit,
+                load_info = self._load_info(
+                    message_count=self._message_count + 1,
+                    total_bytes=self._total_bytes + message.ByteSize(),
                 )
-                error = exceptions.FlowControlLimitError(msg)
-
-                raise error
+                error_msg = "Flow control limits would be exceeded - {}.".format(
+                    load_info
+                )
+                raise exceptions.FlowControlLimitError(error_msg)
 
             assert (
                 self._settings.limit_exceeded_behavior
@@ -124,14 +120,11 @@ class FlowController(object):
                 message.ByteSize() > self._settings.byte_limit
                 or self._settings.message_limit < 1
             ):
-                error_msg = (
-                    "Flow control limits too low for the message. "
-                    "(messages: {} / {}, bytes: {} / {})."
-                ).format(
-                    1,
-                    self._settings.message_limit,
-                    message.ByteSize(),
-                    self._settings.byte_limit,
+                load_info = self._load_info(
+                    message_count=1, total_bytes=message.ByteSize()
+                )
+                error_msg = "Flow control limits too low for the message - {}.".format(
+                    load_info
                 )
                 raise exceptions.PermanentlyBlockedError(error_msg)
 
@@ -145,10 +138,16 @@ class FlowController(object):
                     )
 
                 _LOGGER.debug(
-                    "Blocking until there is enough free capacity in the flow."
+                    "Blocking until there is enough free capacity in the flow - "
+                    "{}.".format(self._load_info())
                 )
+
                 self._has_capacity.wait()
-                _LOGGER.debug("Woke up from waiting on free capacity in the flow.")
+
+                _LOGGER.debug(
+                    "Woke up from waiting on free capacity in the flow - "
+                    "{}.".format(self._load_info())
+                )
 
             # Message accepted, increase the load and remove thread stats if
             # they exist in the waiting queue.
@@ -254,3 +253,41 @@ class FlowController(object):
         msg_count_overflow = self._message_count + 1 > self._settings.message_limit
 
         return size_overflow or msg_count_overflow
+
+    def _load_info(self, message_count=None, total_bytes=None, reserved_bytes=None):
+        """Return the current flow control load information.
+
+        The caller can optionally adjust some of the values to fit its reporting
+        needs.
+
+        The method assumes that the caller has obtained ``_operational_lock``.
+
+        Args:
+            message_count (Optional[int]):
+                The value to override the current message count with.
+            total_bytes (Optional[int]):
+                The value to override the current total bytes with.
+            reserved_bytes (Optional[int]):
+                The value to override the current number of reserved bytes with.
+
+        Returns:
+            str
+        """
+        msg = "messages: {} / {}, bytes: {} / {} (reserved: {})"
+
+        if message_count is None:
+            message_count = self._message_count
+
+        if total_bytes is None:
+            total_bytes = self._total_bytes
+
+        if reserved_bytes is None:
+            reserved_bytes = self._reserved_bytes
+
+        return msg.format(
+            message_count,
+            self._settings.message_limit,
+            total_bytes,
+            self._settings.byte_limit,
+            reserved_bytes,
+        )
--- a/google/cloud/pubsub_v1/publisher/flow_controller.py
+++ b/google/cloud/pubsub_v1/publisher/flow_controller.py
@@ -201,8 +201,16 @@ class FlowController(object):
 
             reservation = self._byte_reservations[thread]
             still_needed = reservation.needed - reservation.reserved
-            can_give = min(still_needed, available)
 
+            # Sanity check for any internal inconsistencies.
+            if still_needed < 0:
+                msg = "Too many bytes reserved: {} / {}".format(
+                    reservation.reserved, reservation.needed
+                )
+                warnings.warn(msg, category=RuntimeWarning)
+                still_needed = 0
+
+            can_give = min(still_needed, available)
             reservation.reserved += can_give
             self._reserved_bytes += can_give
             available -= can_give
--- a/google/cloud/pubsub_v1/publisher/flow_controller.py
+++ b/google/cloud/pubsub_v1/publisher/flow_controller.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections import deque
 import logging
 import threading
 import warnings
@@ -21,6 +22,14 @@ from google.cloud.pubsub_v1.publisher import exceptions
 
 
 _LOGGER = logging.getLogger(__name__)
+
+
+class _QuantityReservation(object):
+    """A (partial) reservation of a quantifiable resource."""
+
+    def __init__(self, reserved, needed):
+        self.reserved = reserved
+        self.needed = needed
 
 
 class FlowController(object):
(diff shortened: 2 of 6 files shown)


google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -333,6 +341,11 @@ def publish(self, topic, data, ordering_key="", **attrs):
 
             pubsub_v1.publisher.exceptions.MessageTooLargeError: If publishing
                 the ``message`` would exceed the max size limit on the backend.
+
+            :exception:`~pubsub_v1.publisher.exceptions.FlowControlLimitError`:
```
kamalaboulhosn, line 345: In Java, we didn't throw an exception, we completed the future with the exception. Perhaps we should do the same here for consistency.

google/cloud/pubsub_v1/publisher/flow_controller.py
```diff
@@ -0,0 +1,145 @@
...
+                self._message_count -= 1
+                self._total_bytes -= message.ByteSize()
+                raise error
+
+            assert (
+                self._settings.limit_exceeded_behavior
+                == types.LimitExceededBehavior.BLOCK
+            )
+
+            while self._is_overflow():
```
kamalaboulhosn, line 98: I'm not sure this is going to have the desired effect. Imagine I set flow control to allow ten messages through and I call publish 30 times. self._message_count will increment to 30. Now, when a message finishes, we won't call self._has_capacity.notify_all() because self._is_overflow() is still true. In fact, we will never call notify_all because the number of outstanding messages will never drop below 20.

pradn: This issue only occurs when the behavior is set to types.LimitExceededBehavior.BLOCK. If it's set to LimitExceededBehavior.ERROR, message_count/total_bytes are not increased if the publish is rejected. (See line 89-90.)

A simple way to fix this is to reduce the message_count/total_bytes before blocking, and to add them again after the wait returns, but before the is_overflow check is done again. This way, we don't wait while using phantom quota.

Perhaps a more elegant way is to change _is_overflow to take the new message's size as a parameter. Then, we can check if the it would overflow before modifying the variables. Then, we'd only need to increment count/size at the end of the function (end of the `with` block.)

google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -372,6 +392,7 @@ def publish(self, topic, data, ordering_key="", **attrs):
 
             # Delegate the publishing to the sequencer.
             future = sequencer.publish(message)
+            future.add_done_callback(on_publish_done)
```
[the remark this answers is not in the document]
plamut: Looking at the [futures code](https://github.com/googleapis/python-pubsub/blob/master/google/cloud/pubsub_v1/futures.py#L147-L186), done callbacks are called both on future success and error (they both call `_trigger()`), thus the message always gets released from the flow controller.

This also holds for already done futures, since [add_done_callback()](https://github.com/googleapis/python-pubsub/blob/master/google/cloud/pubsub_v1/futures.py#L131-L145) just invokes the callback immediately in this case - or am I missing something?

google/cloud/pubsub_v1/publisher/flow_controller.py
```diff
@@ -0,0 +1,145 @@
...
+                _LOGGER.debug(
+                    "Blocking until there is enough free capacity in the flow."
+                )
+                self._has_capacity.wait()
+                _LOGGER.debug("Woke up from waiting on free capacity in the flow.")
+
+    def release(self, message):
+        """Release a mesage from flow control.
```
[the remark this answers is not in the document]
plamut:
```suggestion
        """Release a message from flow control.
```

google/cloud/pubsub_v1/publisher/flow_controller.py
plamut: > Now, when a message finishes, we won't call self._has_capacity.notify_all() because self._is_overflow() is still true. In fact, we will never call notify_all because the number of outstanding messages will never drop below 20.

By design, `notify_all()` acts on a "negative front", i.e. when the state changes from "overflow" to "not overflow anymore". Every release decreases the load, and while a single released message will indeed not call `notify_all()`, the latter will be called after releasing enough messages.

> I'm not sure this is going to have the desired effect. Imagine I set flow control to allow ten messages through and I call publish 30 times. self._message_count will increment to 30.

That's a bug, thanks for catching it! The message count should never increase to 30, but should instead block when 11th message arrives. That message should only be accepted, i.e. added to the load, after releasing at least one message.

Essentially the case covered in the ERROR path (the load adjustment @pradn mentioned), but forgot to add the same to the BLOCK path.

> Perhaps a more elegant way is to change _is_overflow to take the new message's size as a parameter. Then, we can check if the it would overflow before modifying the variables.

Indeed. :+1: Fixing the BLOCK path would result in too many load adjustments scattered around the method, hindering readability, thus the overflow logic will probably be refactored.

google/cloud/pubsub_v1/publisher/client.py
plamut: I'll see if that can be implemented without violating the boundaries too much. Maybe by `add()` also accepting the publish future, and resolving it with an exception rather than raising the exception directly...

... and there you go, answered it to myself. :smile:

google/cloud/pubsub_v1/publisher/flow_controller.py
plamut: @kamalaboulhosn @pradn Thank you both for the valuable comments!

I actually thought we indeed have a bug in the BLOCK path, but thinking through it again makes me believe we got it right after all. At least I was not able to reproduce the supposed buggy behavior with a few additional tests I ran locally, but will nevertheless check the existing FlowController tests once again to see if they somehow missed something.

plamut:
I added support for multithreaded `add()`, including the logic to prevent threads adding large messages from starving.

Things left:
 - [ ] Improve debug log messages (include info on the current flow controller capacity)
 - [ ] ERROR path: raise overflow errors through the publish future, not directly.
- [ ] Add a smoke system test - probably using the ERROR path, as triggering a timeout would be too difficult on reasonably fast connection (we don't want to DOS the infrastructure, after all).

google/cloud/pubsub_v1/publisher/flow_controller.py
plamut: I figured out a simple fix would possibly lead to thread starving - a thread adding a large message would "never" detect enough capacity, since other threads posting smaller message would keep stealing the chunks of available capacity from it.

Solved by by keeping track of waiting threads and greedily distributing any freed-up capacity among them in FIFO order.

google/cloud/pubsub_v1/publisher/flow_controller.py
```diff
@@ -62,46 +83,80 @@ def add(self, message):
...
-                    "Flow control limits exceeded "
+                    "Flow control limits would be exceeded "
                     "(messages: {} / {}, bytes: {} / {})."
                 ).format(
-                    self._message_count,
+                    self._message_count + 1,
                     self._settings.message_limit,
-                    self._total_bytes,
+                    self._total_bytes + message.ByteSize(),
                     self._settings.byte_limit,
```
plamut: Note to self - we probably need to include _reserved_ bytes as well.

tests/unit/pubsub_v1/publisher/test_flow_controller.py
```diff
@@ -0,0 +1,363 @@
...
+        flow_controller.add(msg)
+
+    error_msg = str(error_info.value)
+    assert "messages: 1 / 0" in error_msg
+    assert "bytes: {} / 1".format(msg.ByteSize()) in error_msg
+
+
+def test_threads_posting_large_messages_do_not_starve():
+    settings = types.PublishFlowControl(
+        message_limit=100,  # simulate non-sane settings
```
plamut, line 310:
```suggestion
        message_limit=100,
```
Incorrect comment copied from another test case.

google/cloud/pubsub_v1/publisher/client.py
plamut: Indeed we should, thanks. We do not not include errors raised through the publish futures.

plamut:
@pradn Do you have anything to add here, or are we good to merge?

google/cloud/pubsub_v1/publisher/flow_controller.py
```diff
@@ -0,0 +1,293 @@
...
+                message.ByteSize() > self._settings.byte_limit
+                or self._settings.message_limit < 1
+            ):
+                load_info = self._load_info(
+                    message_count=1, total_bytes=message.ByteSize()
+                )
+                error_msg = "Flow control limits too low for the message - {}.".format(
+                    load_info
+                )
+                raise exceptions.PermanentlyBlockedError(error_msg)
```
pradn, line 129: "WouldPermanentlyBlockError" makes more sense, because it's not already permanently blocked.

Even better, I think we can simply omit the details about blocking and say something more generic: FlowControlMessageTooLargeError

google/cloud/pubsub_v1/publisher/flow_controller.py
```diff
@@ -0,0 +1,293 @@
...
+
+        Returns:
+            bool
+        """
+        if self._waiting:
+            # It's enough to only check the head of the queue, because FIFO
+            # distribution of any free capacity.
+            reservation = self._byte_reservations[self._waiting[0]]
+            return (
+                reservation.reserved >= reservation.needed
```
pradn, line 226: Perhaps warn if reservation.reserved > reservation.needed? This should never be the case and indicates a bug.

google/cloud/pubsub_v1/publisher/flow_controller.py
```diff
@@ -0,0 +1,293 @@
...
+                    "{}.".format(self._load_info())
+                )
+
+            # Message accepted, increase the load and remove thread stats if
+            # they exist in the waiting queue.
+            self._message_count += 1
+            self._total_bytes += message.ByteSize()
+
+            reservation = self._byte_reservations.get(current_thread)
+            if reservation:
```
pradn, line 158: When is it possible for 'reservation' to not exist? It seems like we always create one in line 136 if it doesn't exist. And it's not possible for another part of the code to delete them, either.

google/cloud/pubsub_v1/publisher/flow_controller.py
plamut: Makes sense, I will rename the exception.

google/cloud/pubsub_v1/publisher/flow_controller.py
plamut: It can happen if a thread does not block because there's enough capacity and simply goes through the "would overflow" check - but I will re-check, just in case.

google/cloud/pubsub_v1/publisher/flow_controller.py
plamut: Good idea, especially because we have something similar in line 179 for load stats.

google/cloud/pubsub_v1/publisher/flow_controller.py
pradn: To get to line 133, we must pass the check in line 90, which implies by the time that to get to 133, would_overflow must be true. If it goes in the while loop, we guarantee that a new reservation is made or one already exists. So I don't think this check is necessary.

plamut: I checked again and you are right. I don't exactly remember why the check was added, but it's indeed not needed now, nice catch.

google/cloud/pubsub_v1/publisher/flow_controller.py
plamut: Added the check, albeit at a different place (where reservation is actually modified).

google/cloud/pubsub_v1/publisher/flow_controller.py
plamut: I eventually opted for the generic `FlowControlLimitError` for two reasons:
- `FlowControlMessageTooLargeError` could be confusing, as there already exists a similarly named `MessageTooLargeError` for the messages that would exceed the limits on the backend.
- A message could block permanently for reasons other than its size, e.g. if flow control's max messages is set to 0 or some other non-sensible value. Mentioning the size would actually be misleading in such case, but a generic `FlowControlLimitsError` would semantically still be correct.

google/cloud/pubsub_v1/publisher/flow_controller.py
```diff
@@ -76,12 +76,12 @@ def add(self, message):
...
-                If adding a message exceeds flow control limits and the desired
-                action is :attr:`~google.cloud.pubsub_v1.types.LimitExceededBehavior.ERROR`.
-            :exception:`~pubsub_v1.publisher.exceptions.PermanentlyBlockedError`:
-                If adding a message exceeds total flow control limits and would
-                always overflow on its own, and the desired action is
-                :attr:`~google.cloud.pubsub_v1.types.LimitExceededBehavior.BLOCK`.
+                If adding a message would exceed flow control limits and the desired
+                action is :attr:`~google.cloud.pubsub_v1.types.LimitExceededBehavior.ERROR`,
+                or if a message would always exceed total flow control limits on
+                its own and the desired action is
```
pradn, line 82: How about:

Raised when the desired action is :attr:`~google.cloud.pubsub_v1.types.LimitExceededBehavior.ERROR` and the message would exceed flow control limits, or when the desired action is :attr:`~google.cloud.pubsub_v1.types.LimitExceededBehavior.BLOCK` and the message would block forever against the flow control limits.

google/cloud/pubsub_v1/publisher/flow_controller.py
```diff
@@ -204,8 +201,16 @@ def _distribute_available_bytes(self):
 
             reservation = self._byte_reservations[thread]
             still_needed = reservation.needed - reservation.reserved
-            can_give = min(still_needed, available)
 
+            # Sanity check for any internal inconsistencies.
+            if still_needed < 0:
```
pradn, line 206: I usually assert in these buggy cases. Users notice asserts and we fix them. Otherwise, bugs might lurk for a while if no one checks the logs. It's up to you though.

google/cloud/pubsub_v1/publisher/flow_controller.py
pradn: I agree that it's hard to mention the size being the problem when there's a few other reasons as well.

plamut: Changed, I also think the proposed version is worded better.

google/cloud/pubsub_v1/publisher/flow_controller.py
plamut: Using assert could be an option, but after some thought I would still go with a warning message, as this is the approach already used in leaser, for instance, i.e. log and correct the issue.

Apart from consistency, a warning is probably better suited here, since we have a reasonable chance of recovering from an unexpected state, as opposed to `assert` that would just break immediately (or is just a no-op when Python is run in non-debug mode).

Nevertheless, I will bring up this topic at the next libs meeting to check what other libraries usually do (I know that in BigQuery warnings are used).

pradn: Sounds reasonable to me. :)

plamut:
@pradn If there's nothing else left, please just give an approval to unblock merging, thanks! :)

PR #98, opened 2020-05-09
docs: Changed Python 2 EOL notice from future to present

OrBin:
The library has a Python 2 EOL notice in the docs.
Since the set date has passed, I've changed the phrasing from future to present.

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -21,7 +21,7 @@
 
           <div class="body" role="main">
           	<div class="admonition" id="python2-eol"> 
-          	 On January 1, 2020 this library will no longer support Python 2 on the latest released version. 
+          	 Starting January 1, 2020, this library no longer supports Python 2 on the latest released version. 
           	 Previously released library versions will continue to be available. For more information please
           	 visit <a href="https://cloud.google.com/python/docs/python2-sunset/">Python 2 support on Google Cloud</a>.
           	</div>


plamut:
@OrBin Thanks for the pull request, although the library is currently still compatible with Python 2, because the last Python 2 update actually happened in late April 2020 to include the fixes done before the end of 2019.

@busunkim96 What should be the exact wording across libraries? Technically, we still support Python 2 and we have not yet made a release that is not compatible with it, at least not in PubSub.

Thanks!

busunkim96:
I'd prefer to keep the current wording and change it when the library officially stops working with 2. Open to other ideas though.

plamut:
While far from ideal, the old wording describes the current state of affairs slightly better, IMO.

To me it looks like a neglected future notice added before 2020, while using the past tense would actually make the claim about dropped Python 2 support stronger - which has not come into effect yet.

Let's put this PR on hold until the facts match the new wording better. :)

OrBin:
Sounds good to me. 
Another option is rewording to fit the current state, but I guess that's OK too.
Do you expect to release a new version soon?

plamut:
@OrBin From what I heard, some of the Google Cloud Python client libraries will start dropping Python 2.7 support in the near future, though no concrete day for PubSub yet.

As for the next PubSub release - one minor release was made yesterday, and the next one will likely follow some time after publisher flow control is added (PR #96). The latter will still work with Python 2.7.

plamut:
Closing in favor of #158, the latter will make the exact changes that are consistent with docs changes in other libraries.

PR #103, opened 2020-05-22
fix: PubSub incompatibility with api-core 1.17.0+

### LINKED ISSUE #93: Resolve Incompatibility with API Core >= 1.17.0
Changes in 1.17.0 are causing an incompatibility with pubsub. 

# Related Issues: 

https://github.com/googleapis/python-api-core/issues/25
https://github.com/googleapis/python-pubsub/issues/74

# Suspected changes that caused the behavior change:

https://github.com/googleapis/python-api-core/commit/2b103b60ece16a1e1bc98cfda7ec375191a90f75#diff-d97e81006eaaf29e33270a85aead6aafR146-R148

# Repro:

```

import threading
from time import sleep

import logging

def main():
    logging.basicConfig(level=logging.DEBUG)
    logger = logging.getLogger('sample')
    logger.setLevel(logging.DEBUG)
    PROJECT_ID = "crwilcox-test"
    TOPIC_PATH = "my-topic"
    SUBSCRIPTION_PATH = "my-sub"

    publisher_thread = threading.Thread(target=publisher, args=(PROJECT_ID, TOPIC_PATH))
    publisher_thread.daemon = True
    publisher_thread.start()

    subscriber(PROJECT_ID, SUBSCRIPTION_PATH)

def publisher(
    project_id, topic_name
):
    # TODO project_id = "Your Google Cloud Project ID"
    # TODO topic_name = "Your Pub/Sub topic name"


    # The `topic_path` method creates a fully qualified identifier
    # in the form `projects/{project_id}/topics/{topic_name}`
    topic_path = publisher.topic_path(project_id, topic_name)
    #publisher.create_topic(topic_path)

    for n in range(1, 100000):
        data = u"Message number {}".format(n)
        # Data must be a bytestring
        data = data.encode("utf-8")
        # When you publish a message, the client returns a future.
        future = publisher.publish(topic_path, data=data)
        print(future.result())
        sleep(5)

    print("Published messages.")



def subscriber(
    project_id, subscription_name, timeout=None
):
    # TODO project_id = "Your Google Cloud Project ID"
    # TODO subscription_name = "Your Pub/Sub subscription name"
    # TODO timeout = 5.0  # "How long the subscriber should listen for
    # messages in seconds"

    # The `subscription_path` method creates a fully qualified identifier
    # in the form `projects/{project_id}/subscriptions/{subscription_name}`
    subscription_path = subscriber.subscription_path(
        project_id, subscription_name
    )

    def callback(message):
        print("Received message: {}".format(message.data))
        message.ack()

    streaming_pull_future = subscriber.subscribe(
        subscription_path, callback=callback
    )
    print("Listening for messages on {}..\n".format(subscription_path))

    # Wrap subscriber in a 'with' block to automatically call close() when done.
    with subscriber:
        try:
            # When `timeout` is not set, result() will block indefinitely,
            # unless an exception is encountered first.
            streaming_pull_future.result(timeout=timeout)
        except:  # noqa


main()

```

# Result:

Eventually, subscription will fail and stop processing.

The failure exists in bidi on getting the next item via `__iter__`

```
DEBUG:google.cloud.pubsub_v1.publisher._batch.thread:gRPC Publish took 0.034303903579711914 seconds.
1162512773129262
DEBUG:google.cloud.pubsub_v1.subscriber._protocol.streaming_pull_manager:Sent request(s) over unary RPC.
DEBUG:google.cloud.pubsub_v1.subscriber._protocol.streaming_pull_manager:Released held message, scheduling callback for it, still on hold 0 (bytes 0).
DEBUG:google.api_core.bidi:waiting for recv.
Received message: b'Message number 117'
DEBUG:google.cloud.pubsub_v1.subscriber._protocol.dispatcher:Handling 1 batched requests
DEBUG:google.cloud.pubsub_v1.subscriber._protocol.streaming_pull_manager:Sent request(s) over unary RPC.
iter except: <_MultiThreadedRendezvous of RPC that terminated with:
        status = StatusCode.DEADLINE_EXCEEDED
        details = "Deadline Exceeded"
        debug_error_string = "{"created":"@1588195797.686297000","description":"Error received from peer ipv4:216.58.217.42:443","file":"src/core/lib/surface/call.cc","file_line":1056,"grpc_message":"Deadline Exceeded","grpc_status":4}"
>
INFO:google.cloud.pubsub_v1.subscriber._protocol.streaming_pull_manager:Observed non-terminating stream error 504 Deadline Exceeded
INFO:google.cloud.pubsub_v1.subscriber._protocol.streaming_pull_manager:Observed recoverable stream error 504 Deadline Exceeded
DEBUG:google.api_core.bidi:Re-opening stream from gRPC callback.
DEBUG:google.api_core.bidi:Empty queue and inactive call, exiting request generator.
DEBUG:google.cloud.pubsub_v1.subscriber._protocol.leaser:The current deadline value is 10 seconds.
```

```
PYTHON_PATH/lib/python3.8/threading.py(890)_bootstrap()
-> self._bootstrap_inner()
  PYTHON_PATH/lib/python3.8/threading.py(932)_bootstrap_inner()
-> self.run()
  PYTHON_PATH/lib/python3.8/threading.py(870)run()
-> self._target(*self._args, **self._kwargs)
  /workspace/googleapis/python-pubsub/venv/lib/python3.8/site-packages/google/api_core/bidi.py(655)_thread_main()
-> response = self._bidi_rpc.recv()
  /workspace/googleapis/python-pubsub/venv/lib/python3.8/site-packages/google/api_core/bidi.py(562)recv()
-> return self._recoverable(self._recv)
  /workspace/googleapis/python-pubsub/venv/lib/python3.8/site-packages/google/api_core/bidi.py(505)_recoverable()
-> return method(*args, **kwargs)
  /workspace/googleapis/python-pubsub/venv/lib/python3.8/site-packages/google/api_core/bidi.py(559)_recv()
-> return next(call)
> /workspace/googleapis/python-pubsub/venv/lib/python3.8/site-packages/google/api_core/grpc_helpers.py(109)next()
-> six.raise_from(exceptions.from_grpc_error(exc), exc)
```

plamut:
Fixes #93.

This PR fixes the bug introduced in api-core 1.17.0 that breaks streaming pull recovery on recoverable errors.

The fix depends on the related [pull request](https://github.com/googleapis/python-api-core/pull/30) in API core that allows disabling the automatic pre-fetch of the first stream result.

Probably best to **not merge** until a new `api-core` version is released, as the fix here will not have any effect on its own, and will not close the issue yet.

cc: @pradn 

### PR checklist


plamut:
The docs failure seems to be unrelated to the actual change, it was just revealed now that we re-generated the files from protos.

dhendry:
Please prioritize -  we use pipenv and this compatibility problem has broken our development flow

plamut:
@dhendry As of PubSub 1.4.3, the `api-core` dependency has been restricted to `version < 1.17.0` and should not be experiencing this issue. Or is there another factor involved?

In any case, in order for this fix to work, a [change](https://github.com/googleapis/python-api-core/pull/30) also needs to be done (and released) in api-core. The change got a preliminary approval, and is now just awaiting some related check.

cc: @crwilcox Do you have free cycles to prioritize the check? Thanks!

dhendry:
@plamut We ran into problems because the latest version google-api-python-client (1.9.1) requires `"google-api-core>=1.17.0,<2dev",` which is incompatible with the google-cloud-pubsub restriction of < 1.17.0

For us `google-api-python-client` was being pulled in implicitly as a dependency of `firebase-admin`

plamut:
@dhendry Ah, understood. The `pip` resolver should not allow that to be installed, but it currently has some bugs.

Anyhow, I will push for a new PubSub fix + release as soon as the `api-core` fix is merged.

plamut:
@dhendry As a temporary workaround, would it be feasible to pin and install an older version of `google-api-python-client` that does not require `api-core` >=1.17.0?

plamut:
@pradn Another pair of eyes would be useful for the fix on the PubSub side.

I also removed all non-relevant synth changes from the PR to avoid the docs check issue due to mis-indented generated docstrings.

google/cloud/pubsub_v1/gapic/subscriber_client.py
```diff
@@ -1170,6 +1170,11 @@ def streaming_pull(
                 client_info=self._client_info,
             )
 
+            # Wrappers in api-core should not automatically pre-fetch the first
+            # stream result, as this breaks the stream when re-opening it.
+            # https://github.com/googleapis/python-pubsub/issues/93#issuecomment-630762257
```
pradn, line 1175: This is purely a performance thing yes? It shouldn't have any impact otherwise?

plamut: Not sure if I understand the performance part?

The complementing [fix in api-core](https://github.com/googleapis/python-api-core/pull/30/files#diff-d97e81006eaaf29e33270a85aead6aaf) just checks the flag added here and can be used to disable consuming the first stream item on stream creation.

After applying it, the `api-core` will again start behaving the same as in it was before the 1.17.0 release while still preserving [the fix](https://github.com/googleapis/python-api-core/commit/2b103b60ece16a1e1bc98cfda7ec375191a90f75#diff-d97e81006eaaf29e33270a85aead6aaf) that was added for Firestore, but unfortunately broke PubSub in the process.

pradn: Got it, makes sense. I was wondering what the performance implications of not doing the pre-fetching, but it seems like it's negligible.

plamut:
We need to wait with merging until the `api-core PR` is merged and released, otherwise this fix will have no effect (even worse - it can install one of the more recent `api-core` releases that do not have the required fix yet).

arithmetic1728:
@plamut The api-core PR is merged and just released (v1.20.0).

pradn:
We're going to try to get this in for release 1.6.0.

plamut:
@pradn Please just wait untill I update version pins.

setup.py
```diff
@@ -29,10 +29,10 @@
 # 'Development Status :: 5 - Production/Stable'
 release_status = "Development Status :: 5 - Production/Stable"
 dependencies = [
-    # google-api-core[grpc] 1.17.0 causes problems with stream recovery, thus
-    # it should not be used.
+    # google-api-core[grpc] 1.17.0 up to 1.19.1 causes problems with stream
+    # recovery, thus those versions should not be used.
     # https://github.com/googleapis/python-pubsub/issues/74
-    "google-api-core[grpc] >= 1.14.0, != 1.17.0",
+    "google-api-core[grpc] >= 1.14.0, != 1.17.0, != 1.18.0, != 1.19.0, != 1.19.1",
```
pradn, line 35: Perhaps:
google-api-core[grpc] >= 1.14.0, != 1.17.*, != 1.18.*, != 1.19.0, !=1.19.1
instead? The wildcards guarantee we're not forgetting any minor versions in there.

plamut: AFAIK releases are only incremental and it's highly unlikely that a minor bugfix version, say, 1.17.1 is released in the future?

pradn: Yes, but it's also slightly better for docs - you know for sure that there isn't some minor version that's affected. I know we have that in the comment, but comments can change. This is likely going to be around until we make a brand new version of this library.

plamut:
```suggestion
    "google-api-core[grpc] >= 1.14.0, != 1.17.*, != 1.18.*, != 1.19.*",
```

plamut: Agree, made the list of restricted versions wider, all bugfix versions of problematic minor versions are now excluded, too.

plamut: Also made a quick test, seems like the pins are correct:
```py
In [40]: import pkg_resources
In [41]: versions = ['1.16.0', '1.17.0', '1.18.0', '1.18.3', '1.19.0', '1.19.1', '1.20.0']
In [42]: requirement = pkg_resources.Requirement.parse("google-api-core[grpc] >= 1.14.0, != 1.17.*, != 1.18.*, != 1.19.*")
In [43]: list(requirement.specifier.filter(versions))
Out[43]: ['1.16.0', '1.20.0']

```

PR #108, opened 2020-05-28
refactor: incorporate will_accept() checks into publish()

### LINKED ISSUE #105: unused message arg in Batch.will_accept()
Looks like `message` arg is never used by `will_accept()` method:
https://github.com/googleapis/python-pubsub/blob/cf9e87c80c0771f3fa6ef784a8d76cb760ad37ef/google/cloud/pubsub_v1/publisher/_batch/base.py#L112-L136

Most likely it should be deleted to avoid any confusion, but I suppose it should be done with deprecation. `Batch()` is not something pretending to be manually used by users, but who knows.

@pradn, please, comment

IlyaFaer:
Closes #105


google/cloud/pubsub_v1/publisher/_batch/base.py
```diff
@@ -109,32 +109,6 @@ def status(self):
...
-
-        Args:
-            message (~.pubsub_v1.types.PubsubMessage): The Pub/Sub message.
-
-        Returns:
-            bool: Whether this batch can accept the message.
-        """
-        # If this batch is not accepting messages generally, return False.
-        if self.status != BatchStatus.ACCEPTING_MESSAGES:
-            return False
```
IlyaFaer, line 128: This statement moved into `publish()`, and the second one seems to be already there

tests/unit/pubsub_v1/publisher/batch/test_base.py
```diff
@@ -46,33 +46,3 @@ def test_len():
     assert len(batch) == 0
     batch.publish(types.PubsubMessage(data=b"foo"))
     assert len(batch) == 1
-
-
-def test_will_accept():
```
pradn, line 51: We still need an equivalent of this test. It should go in test_thread.py.

tests/unit/pubsub_v1/publisher/batch/test_thread.py
```diff
@@ -287,20 +287,6 @@ def test_publish_updating_batch_size():
     assert batch.size > 0  # I do not always trust protobuf.
 
 
-def test_publish_not_will_accept():
```
pradn, line 290: We also need this + another test for checking max_bytes is enforced. I know we have some tests below that do that, but they're entangled with other concerns, so it's good to unit-ize these two checks.

IlyaFaer: Hm-m, returning this test back causes a problem, which I didn't notice earlier. Here:
https://github.com/googleapis/python-pubsub/blob/cf9e87c80c0771f3fa6ef784a8d76cb760ad37ef/google/cloud/pubsub_v1/publisher/_batch/thread.py#L355

If there are no messages in the batch yet, new message will be accepted, even in case of `max_messages=0` (overflow). I know, `max_messages=0` is illogical, but that is the test case. I would change this condition to just `not overflow`, as checking for `not self._messages` looks pointless to me. What do you think?

Before this changes `will_accept()` was checking for overflow on a `publish()` start, so this test never touched the condition `not self._messages`.

pradn: Yes, I think we can remove "not self._messages" too. Earlier, we do a check to make sure the message is smaller than the max message size. This means by the time it gets here, if there are no messages in _messages, the first message should not overflow. So, if there are no messages in _messages, this check is redundant, because "not overflow" should be true.

+ @plamut, can you double check my logic?

plamut: > Earlier, we do a check to make sure the message is smaller than the max message size. This means by the time it gets here, if there are no messages in _messages, the first message should not overflow. 

By my reading, the first message can overflow even with sensible batch settings, e.g. if `max_bytes` is set to something smaller than `_SERVER_PUBLISH_MAX_BYTES` that we check above, and the message size falls somewhere in between. Without the `not self._messages` check, such a message would not be added to the batch, but still trigger a commit - of an empty batch.

I presume we want to accept such message and publish it immediately?

After all `max_bytes` is not the `byte_limit` setting from the upcoming publish flow controller, it's just a threshold to determine when to commit the batch.

pradn: Hmm, then it comes down to whether we should allow users to set such a low max_bytes. It is useful when a user's messages are small (<< _SERVER_PUBLISH_MAX_BYTES).

I suppose the principle of allowing at least one message into a batch makes sense, too.

I am convinced. Let's keep the messages=0 check and modify the tests accordingly.

IlyaFaer: Okay, added three tests

tests/unit/pubsub_v1/publisher/batch/test_thread.py
```diff
@@ -287,18 +287,43 @@ def test_publish_updating_batch_size():
...
-    batch = create_batch(topic="topic_foo", max_messages=0)
-    base_request_size = types.PublishRequest(topic="topic_foo").ByteSize()
+def test_publish():
```
pradn, line 290: Can you create a variant of this where max_messages = 0? We want to ensure that at least one message gets in the batch if it's <= _SERVER_PUBLISH_MAX_BYTES.

IlyaFaer: Pushed

pradn:
Thanks @IlyaFaer !

PR #113, opened 2020-06-05
docs: fix update subscription/snapshot/topic samples

### LINKED ISSUE #110: docs: update_subscription and update_snapshot examples are wrong
The [example](https://github.com/googleapis/python-pubsub/blob/cf9e87c80c0771f3fa6ef784a8d76cb760ad37ef/google/cloud/pubsub_v1/gapic/subscriber_client.py#L528) subscription object doesn't have a 'name' field.  Same with [update_snapshot](https://github.com/googleapis/python-pubsub/blob/cf9e87c80c0771f3fa6ef784a8d76cb760ad37ef/google/cloud/pubsub_v1/gapic/subscriber_client.py#L1496).

plamut:
Fixes #110.

This PR fixes two samples in subscriber docstrings. Ideally, this should be addressed in the code generator itself, but until then we can fix this manually on the PubSub side.

I did not re-generate all files due to the current issue with docstring format, thus the PR only includes the relevant changes in the generated code.

### PR checklist

### DIFF
--- a/google/cloud/pubsub_v1/gapic/subscriber_client.py
+++ b/google/cloud/pubsub_v1/gapic/subscriber_client.py
@@ -526,10 +526,8 @@ class SubscriberClient(object):
             >>>
             >>> ack_deadline_seconds = 42
             >>> subscription_name = 'projects/my-project/subscriptions/my-subscription'
-            >>> topic_name = 'projects/my-project/topics/my-topic'
             >>> subscription = {
             ...    'name': subscription_name,
-            ...    'topic': topic_name,
             ...    'ack_deadline_seconds': ack_deadline_seconds,
             ... }
             >>> paths_element = 'ack_deadline_seconds'
@@ -1500,10 +1498,8 @@ class SubscriberClient(object):
             >>> seconds = 123456
             >>> expire_time = {'seconds': seconds}
             >>> snapshot_name = 'projects/my-project/snapshots/my-snapshot'
-            >>> topic_name = 'projects/my-project/topics/my-topic'
             >>> snapshot = {
             ...    'name': snapshot_name,
-            ...    'topic': topic_name,
             ...    'expire_time': expire_time,
             ... }
             >>> paths_element = 'expire_time'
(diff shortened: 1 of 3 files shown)


google/cloud/pubsub_v1/gapic/subscriber_client.py
```diff
@@ -525,7 +525,13 @@ def update_subscription(
             >>> client = pubsub_v1.SubscriberClient()
             >>>
             >>> ack_deadline_seconds = 42
-            >>> subscription = {'ack_deadline_seconds': ack_deadline_seconds}
+            >>> subscription_name = 'projects/my-project/subscriptions/my-subscription'
+            >>> topic_name = 'projects/my-project/topics/my-topic'
+            >>> subscription = {
+            ...    'name': subscription_name,
+            ...    'topic': topic_name,
```
plamut, line 532: I checked the `.proto` file and `Subscription.topic` is a required field, too.

google/cloud/pubsub_v1/gapic/subscriber_client.py
```diff
@@ -1493,7 +1499,13 @@ def update_snapshot(
             >>>
             >>> seconds = 123456
             >>> expire_time = {'seconds': seconds}
-            >>> snapshot = {'expire_time': expire_time}
+            >>> snapshot_name = 'projects/my-project/snapshots/my-snapshot'
+            >>> topic_name = 'projects/my-project/topics/my-topic'
+            >>> snapshot = {
+            ...    'name': snapshot_name,
+            ...    'topic': topic_name,
```
plamut, line 1506: As per `.proto` file, neither `name` nor `topic` are required, but I believe at least the `name` field should be, thus added both fields to the sample, just in case.

google/cloud/pubsub_v1/gapic/subscriber_client.py
pradn: That's a bit misleading. The "required" field in the Subscription docs refers only to creation time. The "topic" field isn't required for UpdateSubscription. If added WITHOUT adding "topic" to the update_mask, it will have no effect. If added WITH adding to the update_mask, it will result in an error because "topic" is immutable.

So, we should remove the "topic" here.

google/cloud/pubsub_v1/gapic/subscriber_client.py
pradn: We should remove "topic" for the same reason as above.

google/cloud/pubsub_v1/gapic/subscriber_client.py
plamut: Thanks for clarifying what "required" actually means, will remove the topic field then.

PR #120, opened 2020-06-11
chore: configure Python test versions in synth.py

### LINKED ISSUE #119: Configure Python test session versions in noxfile.py
With the recent changes in auto-generated code, the Python versions are not set anymore in `noxfile.py`, the latter only contains placeholders.

We need to update `synth.py` to insert these versions automatically whenever the noxfile is re-generated.

plamut:
Closes #119.

This PR makes sure that Python versions in `noxfile.py` are automatically set on every code re-generation.

To test this, run `synthtool locally`. The noxfile should remain the same, i.e. configured Python versions should preserved.

### PR checklist

### DIFF
--- a/google/cloud/pubsub_v1/gapic/publisher_client.py
+++ b/google/cloud/pubsub_v1/gapic/publisher_client.py
@@ -256,8 +256,10 @@ class PublisherClient(object):
         metadata=None,
     ):
         """
-        Creates the given topic with the given name. See the resource name
-        rules.
+        REQUIRED: The complete policy to be applied to the ``resource``. The
+        size of the policy is limited to a few 10s of KB. An empty policy is a
+        valid policy but certain Cloud Platform services (such as Projects)
+        might reject them.
 
         Example:
             >>> from google.cloud import pubsub_v1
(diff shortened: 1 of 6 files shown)


google/cloud/pubsub_v1/gapic/publisher_client.py
```diff
@@ -256,8 +256,10 @@ def create_topic(
         metadata=None,
     ):
         """
-        Creates the given topic with the given name. See the resource name
-        rules.
+        REQUIRED: The complete policy to be applied to the ``resource``. The
```
pradn, line 259: Does not seem relevant?

google/cloud/pubsub_v1/gapic/publisher_client.py
```diff
@@ -269,12 +271,10 @@ def create_topic(
             >>> response = client.create_topic(name)
 
         Args:
-            name (str): Required. The name of the topic. It must have the format
-                ``"projects/{project}/topics/{topic}"``. ``{topic}`` must start with a
-                letter, and contain only letters (``[A-Za-z]``), numbers (``[0-9]``),
-                dashes (``-``), underscores (``_``), periods (``.``), tildes (``~``),
-                plus (``+``) or percent signs (``%``). It must be between 3 and 255
-                characters in length, and it must not start with ``"goog"``.
+            name (str): An annotation that describes a resource reference, see
```
pradn, line 274: These seem to have gotten more general / worse? Here and below.

google/cloud/pubsub_v1/gapic/publisher_client.py
```diff
@@ -822,8 +822,11 @@ def list_topic_snapshots(
             ...         pass
 
         Args:
-            topic (str): Required. The name of the topic that snapshots are attached to.
-                Format is ``projects/{project}/topics/{topic}``.
+            topic (str): Required. User-provided name for this snapshot. If the name is not
```
pradn, line 825: It should be a topic, not a snapshot.

plamut:
@pradn Huh, it seems like the generator has issues, several docstrings ended up at the wrong place. @busunkim96 Is that a known issue?

If this cannot be fixed quickly, I'll ditch the generated changes and only retain the changes to `noxfile.py`, which is the essence of this PR anyway.

busunkim96:
Uh oh. The strange docstring shuffling happened for a while on Autosynth runs, but I've never seen it on a local run. Has this been happening regularly for you?

As for the noxfile removing the Python versions, could you try updating synthtool?

```
python3 -m pip install git+https://github.com/googleapis/synthtool.git
```

I suspect the output here is the result of using the latest templates with an older synthtool version. (Synthtool now pulls the templates from GitHub).

The intent is that the noxfile use the default versions if none are passed. See [code](https://github.com/googleapis/synthtool/blob/a0ee80e0492a03b9b2bfefb5cca7eaf17bf10438/synthtool/gcp/common.py#L77-L8) and [template](https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/python_library/noxfile.py.j2#L29-L31).

If you do ever want to customize the versions you can pass them into `py_library`.

```
templated_files = common.py_library(
    cov_level=100,
    unit_test_python_versions=["3.6", "3.7", "3.8"],
    system_test_python_versions=["3.7"],
)
```

plamut:
Updating the synthtool fixed the problem with versions in noxfile. Since that was the essence of this PR, we can close it (what's left are just some docstring changes, but that's probably too insignificant to warrant a regen and possibly yet another release).

PR #123, opened 2020-06-15
docs: explain how to nack a sync pull message

### LINKED ISSUE #122: Document how to nack synchronous pull messages
As discussed some time ago, there is no explicit `nack()` method on the subscriber client and it might not be immediately obvious how to nack synchronous pull messages. We should explain this in the docs.

plamut:
Closes #122.

This PR adds a short explanation on how to NACK sync pull messages.

### PR checklist

### DIFF
--- a/docs/subscriber/index.rst
+++ b/docs/subscriber/index.rst
@@ -64,13 +64,13 @@ To pull the messages synchronously, use the client's
     subscriber.acknowledge(subscription_path, ack_ids)
 
 The method returns a :class:`~.pubsub_v1.types.PullResponse` instance that
-cointains a list of received :class:`~.pubsub_v1.types.ReceivedMessage`
+contains a list of received :class:`~.pubsub_v1.types.ReceivedMessage`
 instances.
 
 If you want to **nack** some of the received messages (see :ref:`explaining-ack` below),
 you can use the :meth:`~.pubsub_v1.subscriber.client.Client.modify_ack_deadline`
 method and set their acknowledge deadlines to zero. This will cause them to
-expire instantly and the backend will try to re-deliver them.
+be dropped by this client and the backend will try to re-deliver them.
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


docs/subscriber/index.rst
```diff
@@ -67,6 +67,17 @@ The method returns a :class:`~.pubsub_v1.types.PullResponse` instance that
 cointains a list of received :class:`~.pubsub_v1.types.ReceivedMessage`
 instances.
 
+If you want to **nack** some of the received messages (see :ref:`explaining-ack` below),
+you can use the :meth:`~.pubsub_v1.subscriber.client.Client.modify_ack_deadline`
+method and set their acknowledge deadlines to zero. This will cause them to
+expire instantly and the backend will try to re-deliver them.
```
pradn, line 73: "expire instantly" -> "be dropped in this client"

"expire" feels like it would happen in the server, so I want to clarify. I realize this interpretation conflicts with "the backend will try to re-deliver them" but I want to be super clear.

docs/subscriber/index.rst
```diff
@@ -67,6 +67,17 @@ The method returns a :class:`~.pubsub_v1.types.PullResponse` instance that
 cointains a list of received :class:`~.pubsub_v1.types.ReceivedMessage`
```
pradn, line 67: let's fix "cointains" while we're at it?

docs/subscriber/index.rst
plamut: I always thought the "expiration" actually happens on the backend? The client marks a message as "expired" on the backend, and the latter hence schedules it for re-delivery?

Or do you mean that it should be more clear that the "expiration" is triggered by an explicit client action, rather than some server-side event? Also, would perhaps the wording _"be dropped **by** this client"_ be better?
(I read the "in" as if the dropping happens in the client itself, but is not necessarily connected to the state changes on the backend)

My apologies if I get some of the nuances of the English language wrong. :stuck_out_tongue:

docs/subscriber/index.rst
plamut: Sure, why not, will do it.

docs/subscriber/index.rst
pradn: We try not to use the word "expire" unless we're talking about subscriptions expiring due to inactivity. 

"be dropped by this client" is better - it implies clearly that the client is doing it. I want to be clear about what happens in the client vs the backend.

PR #143, opened 2020-07-06
feat: Add support for server-side flow control

kamalaboulhosn:


google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -575,6 +575,8 @@ def _get_initial_request(self, stream_ack_deadline_seconds):
             stream_ack_deadline_seconds=stream_ack_deadline_seconds,
             subscription=self._subscription,
             client_id=self._client_id,
+            max_outstanding_messages=self._flow_control.max_messages,
```
pradn, line 578: I notice there's not any specific feature flag checking in the backend, so this should be safe to set for all requests.

kamalaboulhosn: There is actually a feature flag guarding this behavior and for now, a visibility label as well.

pradn: Yes, I missed the FF-guard. Sending along these fields will be no-ops if the FF is off, so it should be okay.

PR #149, opened 2020-07-09
feat: Add OpenTelemetry integration

### LINKED ISSUE #124: PubSub add OpenTelemetry support
OpenTelemetry should be added to PubSub to provide OpenTracing and OpenCensus functionality as one API and SDK. It would be helpful to use it to trace a message throughout its life cycle.

sethmaxwl:
Closes #124 
Requires. #158.

This PR adds OpenTelemetry tracing to publisher and subscriber clients.
This PR does not interfere with any underlying client functions. OpenTelemetry is an optional dependency that creates a trace that tracks a message from when it is published to when it is received by a subscriber.

### Ongoing concerns
- [x] Adding OpenTelemetry to unit tests as external dependencies.

### PR checklist
- [x]  Make sure to open an issue as a bug/issue before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea


plamut:
This requires dropping Python 2 support, right? (#131)

sethmaxwl:
> This requires dropping Python 2 support, right? (#131)

Yes, OpenTelemetry does not support any Python version before 3.4.

google/cloud/pubsub_v1/opentelemetry_tracing.py
```diff
@@ -0,0 +1,91 @@
...
+except ImportError:
+    _LOGGER.info(
+        "This service supports OpenTelemetry, but OpenTelemetry could"
+        "not be imported. To use OpenTelemetry, please install the"
+        "opentelemetry-api, opentelemetry-sdk, and opentelemetry-instrumentation"
+        "pip modules. See also"
+        "https://opentelemetry-python.readthedocs.io/en/stable/getting-started.html"
+    )
+    USE_OPENTELEMETRY = False
+    pass
```
plamut, line 41: (nit) `pass` is redundant here

google/cloud/pubsub_v1/opentelemetry_tracing.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+@contextmanager
+def create_span(span_name, attributes=None, parent=None):
+    """ Creates a new span
+
+    Args:
+        span_name (str): the name of the new span
+        attributes ([dict], optional): A dictionary
```
plamut, line 50: (nit)
Following the `mypy` notation, optional types are specified as `Optional[typename]`

Applies to multiple places.

google/cloud/pubsub_v1/opentelemetry_tracing.py
```diff
@@ -0,0 +1,91 @@
...
+        "pip modules. See also"
+        "https://opentelemetry-python.readthedocs.io/en/stable/getting-started.html"
+    )
+    USE_OPENTELEMETRY = False
+    pass
+
+
+@contextmanager
+def create_span(span_name, attributes=None, parent=None):
+    """ Creates a new span
```
plamut, line 46: (nit)
Docstring summary lines should use imperative mood.

google/cloud/pubsub_v1/opentelemetry_tracing.py
```diff
@@ -0,0 +1,91 @@
...
+
+    Args:
+        span_name (str): the name of the new span
+        attributes ([dict], optional): A dictionary
+            containing all attributes to add to a span. Defaults to None.
+        parent ([dict], optional): A dictionary
+            containing the attributes of a . Defaults to None.
```
plamut, line 53: A word seems to be missing in this sentence?

google/cloud/pubsub_v1/opentelemetry_tracing.py
```diff
@@ -0,0 +1,91 @@
...
+    if parent is not None:
+        # Form the parent's context from the parent dict provided
+        try:
+            parent_span_context = SpanContext(
+                trace_id=parent["trace_id"],
+                span_id=parent["span_id"],
+                is_remote=parent["is_remote"],
+                trace_flags=parent["trace_flags"],
+                trace_state=parent["trace_state"],
+            )
```
plamut, line 76: Since `parent` is a mapping, it can be unpacked directly:
```py
parent_span_context = SpanContext(**parent)
```

google/cloud/pubsub_v1/opentelemetry_tracing.py
```diff
@@ -0,0 +1,91 @@
...
+        except:
+            parent_span_context = None
```
plamut, line 78: What exceptions are we catching here? It's advisable to be as specific as possible, since as written, the code will silently swallow even typos in variable names and similar unrelated errors.

google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -369,38 +372,47 @@ def publish(self, topic, data, ordering_key="", **attrs):
...
-        # Create the Pub/Sub message object.
-        message = types.PubsubMessage(
-            data=data, ordering_key=ordering_key, attributes=attrs
-        )
+        span_name = "{} publisher".format(topic)
+        span_attributes = {"data": data.decode()}
+        with create_span(span_name, attributes=span_attributes) as span:
+            if span is not None:
+                # Add the context of the span as an attribute
+                attrs["googclient_OpenTelemetrySpanContext"] = json.dumps(
```
plamut, line 380: Sanity check - what if somebody explicitly sets a message attribute named `googclient_OpenTelemetrySpanContext`?

It's probably quite unlikely, but it probably makes sense to at least issue a warning that this message attribute will be overridden.

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -619,20 +621,40 @@ def _on_response(self, response):
...
-                self._on_hold_bytes += message.size
-                req = requests.LeaseRequest(
-                    ack_id=message.ack_id,
-                    byte_size=message.size,
-                    ordering_key=message.ordering_key,
-                )
-                self.leaser.add([req])
+                if (
+                    "googclient_OpenTelemetrySpanContext"
+                    in received_message.message.attributes.keys()
```
plamut, line 626: Dicts support direct containment check, i.e. `"foo" in my_dict`, we do not need to obtain the keys first with the `keys()` method.

noxfile.py
```diff
@@ -70,7 +70,14 @@ def lint_setup_py(session):
...
 def default(session):
     # Install all test dependencies, then install this package in-place.
-    session.install("mock", "pytest", "pytest-cov")
+    session.install(
+        "mock",
+        "pytest",
+        "pytest-cov",
+        "opentelemetry-api",
+        "opentelemetry-sdk",
+        "opentelemetry-instrumentation",
```
plamut, line 79: This change will get wiped out on the next code regen, let's add dependencies through the `synth.py` file instead to make them persistent.

tests/unit/pubsub_v1/opentelemetry/test_opentelemetry.py
```diff
@@ -0,0 +1,97 @@
...
+    return {
+        "trace_id": context.trace_id,
+        "span_id": context.span_id,
+        "is_remote": context.is_remote,
+        "trace_flags": context.trace_flags,
+        "trace_state": context.trace_state,
+    }
+
+
+class TestOpenTelemetry(unittest.TestCase):
```
plamut, line 60: Recommendation - most (all?) of the existing unit tests use the `pytest` framework, thus let's be consistent and use it in these tests, too. Should make them more concise with less boilerplate. :wink:

tests/unit/pubsub_v1/publisher/test_publisher_client.py
```diff
@@ -28,6 +29,11 @@
 from google.cloud.pubsub_v1.publisher import exceptions
 from google.cloud.pubsub_v1.publisher._sequencer import ordered_sequencer
 
+import opentelemetry
+
+# Disable OpenTelemetry for publisher tests
+sys.modules["opentelemetry"] = None
```
plamut, line 35: Caution - this modifies the global state, i.e. the `sys.modules` dictionary, and the change is not restored after the publisher tests are run. Let's make sure that the fixture is undone when not needed anymore.

plamut (commented):
Disclaimer - I would have to familiarize myself with the `opentelemetry` library more, thus I primarily focused on the code aspects, and less on the actual semantics. Added suggestions at places where the code could benefit from them.

noxfile.py
sethmaxwl: Adding these as external dependencies may not work since there is not a synthtool option to add external unit test dependencies, although there is an option for external system test dependencies. Should these tests be added as system tests?

tests/unit/pubsub_v1/opentelemetry/test_opentelemetry.py
sethmaxwl: I used the unittest framework here because of the setup and teardown methods that make it easier to create the OpenTelemetry environment for the test cases. This can be done for each test in the `pytest` framework, but there would be some repeated code.

plamut: I'm curious where `@pytest.fixture` falls short here? In should be possible to achieve the same setup / teardown behavior without code repetition and just pass the fixture as an argument to the tests that need it.

I'm not super strongly opinionated here, though, the unittest framework does the job fine, it's just that using a single test framework means somewhat lower maintenance cost in the long run.

noxfile.py
plamut: It's not ideal, but we can circumvent this limitation and add additional dependencies through the `synth.py` file. That way the changes to `noxfile.py` will not get lost on the next code regen.

Applying the following patch should do the trick:

```patch
diff --git noxfile.py noxfile.py
index 3af3d67..20b5ad2 100644
--- noxfile.py
+++ noxfile.py
@@ -70,14 +70,12 @@ def lint_setup_py(session):
 
 def default(session):
     # Install all test dependencies, then install this package in-place.
+
+    session.install("mock", "pytest", "pytest-cov")
     session.install(
-        "mock",
-        "pytest",
-        "pytest-cov",
-        "opentelemetry-api",
-        "opentelemetry-sdk",
-        "opentelemetry-instrumentation",
+        "opentelemetry-api", "opentelemetry-sdk", "opentelemetry-instrumentation",
     )
+
     session.install("-e", ".")
 
     # Run py.test against the unit tests.
diff --git synth.py synth.py
index 0e2c96e..10edae8 100644
--- synth.py
+++ synth.py
@@ -279,4 +279,18 @@ s.move(templated_files)
 # ----------------------------------------------------------------------------
 python.py_samples()
 
+# ----------------------------------------------------------------------------
+# Additional unit test dependencies
+# ----------------------------------------------------------------------------
+s.replace(
+    "noxfile.py",
+    r'session\.install\("mock", "pytest", "pytest-cov"\)',
+    """
+    \g<0>
+    session.install(
+        "opentelemetry-api", "opentelemetry-sdk", "opentelemetry-instrumentation",
+    )
+    """,
+)
+
 s.shell.run(["nox", "-s", "blacken"], hide_output=False)
```

sethmaxwl:
> Disclaimer - I would have to familiarize myself with the `opentelemetry` library more, thus I primarily focused on the code aspects, and less on the actual semantics. Added suggestions at places where the code could benefit from them.

All suggestions have been addressed.

plamut (commented):
LGTM now, although we might have to do some additional changes once [#158](https://github.com/googleapis/python-pubsub/pull/158) is merged.

If you have time, you might want to rebase your branch onto that PR branch and see if things work there, too.

google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -369,38 +372,54 @@ def publish(self, topic, data, ordering_key="", **attrs):
...
-            self._flow_controller.add(message)
-        except exceptions.FlowControlLimitError as exc:
-            future = futures.Future()
-            future.set_exception(exc)
-            return future
+                if "googclient_OpenTelemetrySpanContext" in attrs:
+                    _LOGGER.warn(
+                        "googclient_OpenTelemetrySpanContext set on message"
+                        "as an attribute, but will be overridden."
+                    )
```
plamut, line 384: Oh, forgot about this - I don't think we test this code path yet, i.e. whether a warning is indeed emitted.

README.rst
```diff
@@ -186,14 +186,39 @@ For example, to use JSON Web Tokens, provide a `google.auth.jwt.Credentials`_ in
...
+.. code-block:: python
+
+    from opentelemetry import trace
+    from opentelemetry.sdk.trace import TracerProvider
+    from opentelemetry.sdk.trace.export import SimpleExportSpanProcessor
+    from opentelemetry.exporter.cloud_trace import CloudTraceSpanExporter
+
+    trace.set_tracer_provider(TracerProvider())
+    trace.get_tracer_provider().add_span_processor(
+        SimpleExportSpanProcessor(CloudTraceSpanExporter())
```
AndrewAXue, line 215: You should be using a `BatchSpanProcessor` here. `SimpleExportSpanProcessor` exports all spans sequentially and will be very slow.

google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -369,38 +372,54 @@ def publish(self, topic, data, ordering_key="", **attrs):
...
-        try:
-            self._flow_controller.add(message)
-        except exceptions.FlowControlLimitError as exc:
-            future = futures.Future()
-            future.set_exception(exc)
-            return future
+                if "googclient_OpenTelemetrySpanContext" in attrs:
+                    _LOGGER.warning(
+                        "googclient_OpenTelemetrySpanContext set on message"
+                        "as an attribute, but will be overridden."
```
chrisrossi, line 383: No space between "message" and "as".

google/cloud/pubsub_v1/opentelemetry_tracing.py
```diff
@@ -0,0 +1,91 @@
...
+
+    USE_OPENTELEMETRY = True
+except ImportError:
+    _LOGGER.info(
+        "This service supports OpenTelemetry, but OpenTelemetry could"
+        "not be imported. To use OpenTelemetry, please install the"
+        "opentelemetry-api and opentelemetry-instrumentation"
+        "pip modules. See also"
+        "https://opentelemetry-python.readthedocs.io/en/stable/getting-started.html"
+    )
```
cguardia, line 39: There are no spaces at the end of these lines, which will cause words to be displayed together (e.g. couldnot, rather than could not).

google/cloud/pubsub_v1/opentelemetry_tracing.py
```diff
@@ -0,0 +1,91 @@
...
+    if parent is not None:
+        # Form the parent's context from the parent dict provided
+        try:
+            parent_span_context = SpanContext(**parent)
+        except TypeError:
+            _LOGGER.warning(
+                "A parent span was provided but it could not be"
+                "converted into a SpanContext. Ensure that the"
+                "parent is a mapping with at least a trace_id, span_id"
+                "and is_remote keys."
```
cguardia, line 76: Same issue from above, where spaces are missing at the end of the lines.

synth.py
```diff
@@ -279,4 +279,17 @@ def _merge_dict(d1, d2):
 # ----------------------------------------------------------------------------
 python.py_samples()
 
+# ----------------------------------------------------------------------------
+# Additional unit test dependincies
```
cguardia, line 283: Typo: dependencies.

google/cloud/pubsub_v1/opentelemetry_tracing.py
```diff
@@ -0,0 +1,91 @@
...
+    USE_OPENTELEMETRY = True
+except ImportError:
+    _LOGGER.info(
+        "This service supports OpenTelemetry, but OpenTelemetry could"
+        "not be imported. To use OpenTelemetry, please install the"
+        "opentelemetry-api and opentelemetry-instrumentation"
+        "pip modules. See also"
+        "https://opentelemetry-python.readthedocs.io/en/stable/getting-started.html"
+    )
+    USE_OPENTELEMETRY = False
```
cguardia, line 40: Just a suggestion: In addition to the import try, wouldn't it be possible to use an environment variable for explicitly disabling opentelemetry? That way the tests below would not need to mess with sys.modules.

cguardia (approved):
There are still a couple of suggestions from previous reviews, though they are mostly documentation and log warning formatting issues.

cguardia (changes requested):
Seems like lint is not passing, plus there are a few failing tests. Please review.

plamut:
Side note - before merging this (after releasing a new major version, of course), we must also run a few benchmarks to verify that the tracing does not hinder the performance, especially throughput. Mentioning as a a reminder to ourselves.

johnbryan:
Hi all, Seth (author of this PR) finished his Google internship in mid-August so I assume is not actively maintaining this PR now. I would be willing to address the remaining comments here but will not be able to get to it for a couple of weeks. Seth if I'm mistaken and you are indeed still following/working on this let us know, but I assume you have other things to do!

plamut:
@johnbryan That sounds good, thanks!

No rush, though, as we need too release a new major version first anyway.

plamut:
@johnbryan FYI, a new major version has been released, unblocking this PR (which needs an update now, however).

anguillanneuf:
@johnbryan Will you have a chance to rebase this PR?

pamarc:
Hi, any news on this PR? This could be of use :)

plamut:
Just as a quick status update, it is still not clear what exactly do want to be tracked, still awaiting the official specs. We can thus consider this PR blocked until further notice.

CyberHippo:
Hi, is there any update/ETA on this PR?  :)

dmallanb:
This PR has been open a while - any updates as to why it's still blocked?

pedroserrudo (approved):

google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -369,38 +372,54 @@ def publish(self, topic, data, ordering_key="", **attrs):
                 "be sent as text strings."
             )
 
-        # Create the Pub/Sub message object.
-        message = types.PubsubMessage(
-            data=data, ordering_key=ordering_key, attributes=attrs
-        )
+        span_name = "{} publisher".format(topic)
+        span_attributes = {"data": data.decode()}
```
dmaier-FN, line 376: What is the argument for doing this? It seems to me that, _best_ case, it results in the duplication of data in the message. But in the _worst_ case it actually leaks sensitive information into observability stack, just like logging the output of a SQL query would.

dmallanb: I've also noticed it's a potentially unneeded change after experimenting with alternatives myself. Trace information can be passed into pubsub attributes in the current iteration of the pubsub library using Open Telemetry manual instrumentation as far as I can see for now.

mukund-ananthu:
Closing this, since its now done.

PR #152, opened 2020-07-10
samples: add subscription detachment sample

anguillanneuf:
Add a sample for subscription detachment. The test can't be run yet because the feature is still being rolled out.

**The service is live, but we may want to hold off publishing this sample until the library has transitioned to using the microgenerator.**


samples/snippets/iam_test.py
```diff
@@ -110,9 +110,17 @@ def test_check_topic_permissions(topic, capsys):
...
+    iam.check_subscription_permissions(PROJECT_ID, SUBSCRIPTION_ID)
 
     out, _ = capsys.readouterr()
 
     assert subscription in out
     assert "pubsub.subscriptions.consume" in out
+
+
+def test_detach_subscription(subscription, capsys):
+    iam.check_subscription_permissions(PROJECT_ID, SUBSCRIPTION_ID)
```
pradn, line 122: The pubsub.topics.detachSubscription permission is on the topic, not the subscription actually.

anguillanneuf: Ooh, I forgot to update here. Fixed.

samples/snippets/iam_test.py
```diff
@@ -119,7 +119,7 @@ def test_check_subscription_permissions(subscription, capsys):
 
 
 def test_detach_subscription(subscription, capsys):
-    iam.check_subscription_permissions(PROJECT_ID, SUBSCRIPTION_ID)
+    iam.detach_subscription(PROJECT_ID, SUBSCRIPTION_ID)
```
pradn, line 122: This method isn't available IAM. We should write this test like this:
```
def test_check_detach_subscription_permission(topic, capsys):
    iam.check_topic_permissions(PROJECT_ID, TOPIC_ID)	

    out, _ = capsys.readouterr()

    assert topic in out
    assert "pubsub.topics.detachSubscription" in out
```

anguillanneuf: iam refers to `/samples/snippets/iam.py`. This is a unit test for the sample only.

pradn: Oh! Sorry, I thought it was an imported IAM library when I saw the imports in this file.

PR #157, opened 2020-07-14
Revert "chore(deps): update dependency google-cloud-pubsub to v1.7.0"

anguillanneuf:
Reverts googleapis/python-pubsub#155


plamut:
I set the "do not merge" label on the original PR to not interfere with the upcoming major version bump, as we will have to deal with any problems there anyway.

PR #170, opened 2020-07-24
fix: convert all RPC error types to exceptions

### LINKED ISSUE #163: TypeError: exceptions must derive from BaseException
using the subscriber client api a TypeError will often get thrown from `pubsub_v1/futures.py`.  Consider the following:

```
subscription_path = subscriber.subscription_path(CONFIG.GCP_PROJECT, CONFIG.SUBSCRIPTION_NAME)

flow_control = pubsub_v1.types.FlowControl(max_bytes=100 * 1024 * 1024, max_messages=1, max_lease_duration=60 * 10)

future = subscriber.subscribe(
    subscription_path,
    callback=self._tornado_sync_wrapper,
    flow_control=flow_control,
    scheduler=self._make_single_threaded_scheduler(),
)

try:
    future.result()
except Exception as ex:
    logger.error("Exception while trying to run queue listener", exc_info=True)
    future.cancel()
```

this bit of code in pubsub_v1/futures.py can return a TypeError `exceptions must derive from BaseException`:
```
    def result(self, timeout=None):
        """Resolve the future and return a value where appropriate.

        Args:
            timeout (Union[int, float]): The number of seconds before this call
                times out and raises TimeoutError.

        Raises:
            concurrent.futures.TimeoutError: If the request times out.
            Exception: For undefined exceptions in the underlying
                call execution.
        """
        # Attempt to get the exception if there is one.
        # If there is not one, then we know everything worked, and we can
        # return an appropriate value.
        err = self.exception(timeout=timeout)
        if err is None:
            return self._result
        raise err

    def exception(self, timeout=None):
        """Return the exception raised by the call, if any.

        Args:
            timeout (Union[int, float]): The number of seconds before this call
                times out and raises TimeoutError.

        Raises:
            concurrent.futures.TimeoutError: If the request times out.

        Returns:
            Exception: The exception raised by the call, if any.
        """
        # Wait until the future is done.
        if not self._completed.wait(timeout=timeout):
            raise exceptions.TimeoutError("Timed out waiting for result.")

        # If the batch completed successfully, this should return None.
        if self._result != self._SENTINEL:
            return None

        # Okay, this batch had an error; this should return it.
        return self._exception
```

The type returned from `self.exception` can be of type `<opencensus.ext.grpc.utils.WrappedResponseIterator object>` which is not of type Exception.

```
TypeError: exceptions must derive from BaseException
  File "triggers/trigger_pushd.py", line 186, in run_listener
    future.result()
  File "/usr/local/lib/python3.7/dist-packages/google/cloud/pubsub_v1/futures.py", line 107, in result
    raise err
```

#### Environment details

  - OS type and version: Ubuntu 18.04
  - Python version: Python 3.7
  - pip version: 9.0.1
  - `google-cloud-pubsub` version: 1.7.0

plamut:
Fixes #163.

This PR fixes a problem when the underlying gRPC channel is wrapped into something and terminates with a reason that is not an instance of `grpc.RpcError`. The streaming pull manager now converts all reason types into an `Exception` instance, fixing a `TypeError` when a `StreamingPullFuture` then tries to raise it.

Still waiting for the reporter's confirmation that this is indeed what happens, but I found no other place in the code that could call `future.set_exception(...)` with a non-Exception argument.

(probably also makes sense to merge #158 first so that we don't have to make yet another release before dropping Python 2)

### PR checklist


pradn:
Looks good, thanks!

PR #187, opened 2020-09-10
test: fix flaky sequencer unit tests

### LINKED ISSUE #178: Flaky tests
These two are pretty consistently flaky.

test_publisher_client.py::test_wait_and_commit_sequencers
test_publisher_client.py::test_stopped_client_does_not_commit_sequencers

plamut:
Fixes #178.

Patching the client under test should be done on an instance used in a test, not on the instance's class - patching the latter can cause all other instances of the same class to share the patched method, possibly interfering with the patched method's call count.

This can probably be shipped along the new major version, as it does not interfere with the changes made there.

#### PR checklist

### DIFF
--- a/tests/unit/pubsub_v1/publisher/test_publisher_client.py
+++ b/tests/unit/pubsub_v1/publisher/test_publisher_client.py
@@ -435,9 +435,7 @@ def test_wait_and_commit_sequencers():
 
     # Mock out time so no sleep is actually done.
     with mock.patch.object(time, "sleep"):
-        with mock.patch.object(
-            publisher.Client, "_commit_sequencers"
-        ) as _commit_sequencers:
+        with mock.patch.object(client, "_commit_sequencers") as _commit_sequencers:
             assert (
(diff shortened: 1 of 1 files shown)


pradn (approved):
Ah! This makes sense. I assumed the patch would be reset between tests, but I guess there's no reason why that should be. Thanks for the fix!

plamut:
> I assumed the patch would be reset between tests, but I guess there's no reason why that should be.

Actually, the patch _is_ reset between the tests (that's its purpose after all), but the problem is that patching a method on a class affects all of its instances.

The issue here was that some of the other tests create their own `Thread-PubSubBatchCommitter` threads that invoke the `_commit_sequencers()` method, but if that method is patched on the class, that mocked method is shared between tests. That occasionally messed up the call count when some of those `Thread-PubSubBatchCommitter` threads were still running.

pradn:
I see: patching a class is modifying "global state", which leads to conflicts when multiple threads are using that class-level patched method. Thank you for the explanation!

PR #208, opened 2020-09-24
fix: make fixup script consistent with migration docs

### LINKED ISSUE #199: fixup_pubsub_v1_keywords.py generates incorrect output in some cases
I just tried the fixup script and ran into a few issues.

1. It generates this change even though `PublisherClient.publish` does not take the new style parameters:
```diff
-        publisher_client.publish(topic, bytes(json.dumps(message), 'utf-8'))
+        publisher_client.publish(request = {'topic': topic, 'messages': bytes(json.dumps(message), 'utf-8')})
```
2. It generates this change even though `request` and keyword args are documented to be mutually exclusive:
```diff
-        response = subscriber.pull(subscription_path, max_messages=NUM_MESSAGES, timeout=TIMEOUT)
+        response = subscriber.pull(
+            request = {'subscription': subscription_path, 'max_messages': NUM_MESSAGES}, timeout=TIMEOUT,
+        )
```
3. all of the generated `request` keyword arguments break pep8 (they should not have spaces around `=`)

Stylistically, I find the keyword arguments better to read than all of the inline dicts for the `request` syntax.  It'd be great if the tool had an option to emit the keyword args version.

cguardia:

Fixes #199🦕

The following changes to the fixup script are part of the PR:
- Generate code consistent with the upgrade documentation.
- Add an option to generate the code using keyword arguments instead of a request.
- Generate stylistically consistent code (no spaces in keywords, double quotes for strings).
- Reformat the script itself to use the same code styling.


anguillanneuf (approved):
Thank you for this PR.

PR #212, opened 2020-09-29
README: Fix new method signatures using the request parameter.

Natim:


Natim:
@googlebot I signed it!

README.rst
```diff
@@ -138,8 +138,7 @@ the topic, and subscribe to that, passing a callback function.
         project_id=os.getenv('GOOGLE_CLOUD_PROJECT'),
         sub='MY_SUBSCRIPTION_NAME',  # Set this to something appropriate.
     )
-    subscriber.create_subscription(
-        name=subscription_name, topic=topic_name)
+    subscriber.create_subscription(request={"name": subscription_name, "topic": topic_name})
```
plamut, line 141: Same thing here - `create_subscription()` supports subscription name and topic name as keyword arguments ([source](https://github.com/googleapis/python-pubsub/blob/7228f6c9a4c050bf22bb4bc3582b89b04eaa8702/google/pubsub_v1/services/subscriber/client.py#L298-L309)).

README.rst
```diff
@@ -110,7 +110,7 @@ messages to it
         project_id=os.getenv('GOOGLE_CLOUD_PROJECT'),
         topic='MY_TOPIC_NAME',  # Set this to something appropriate.
     )
-    publisher.create_topic(topic_name)
+    publisher.create_topic(request={"name": topic_name})
```
plamut, line 113: The signature here is fine - this particular method supports subscription name and topic name as keyword arguments ([source](https://github.com/googleapis/python-pubsub/blob/7228f6c9a4c050bf22bb4bc3582b89b04eaa8702/google/pubsub_v1/services/publisher/client.py#L277-L285)).

plamut (commented):
Thank you for your kind pull request!

It does appear, though, that the example in README does not have to be fixed - several methods can be called with kwargs only, it is not necessary to use the `request` argument.

More [details](https://github.com/googleapis/python-pubsub/blob/7228f6c9a4c050bf22bb4bc3582b89b04eaa8702/UPGRADING.md#more-details) can be found in the UPGRADING guide.

Natim:
> More details can be found in the UPGRADING guide.

Ok so I just need to name the parameters. Thank you for the precisions.

PR #225, opened 2020-10-21
chore: re-run synthtool

### LINKED ISSUE #224: handwritten layer didn't handle mtls correctly
The auto-generated layer (google/pubsub_v1) has [mtls logic](https://github.com/googleapis/python-pubsub/blob/master/google/pubsub_v1/services/publisher/client.py#L211-L247). This client creates a mtls channel based on environment variable settings and existence of device certificate.

However, the handwritten layer (google/cloud/pubsub_v1) [creates a channel](https://github.com/googleapis/python-pubsub/blob/master/google/cloud/pubsub_v1/publisher/client.py#L149) itself with mtls logic, so the mtls feature is completely missing.

arithmetic1728:
re-run synthtool to update auto-generated layer.

The latest auto-generated client exposes ssl credentials from its transport, which will be used in the future PR to fix #224


anguillanneuf (approved):
LGTM.

PR #231, opened 2020-11-05
feat: Enable server side flow control by default with the option to turn it off

fayssalmartanigcp:
This change enables sending flow control settings automatically to the server. If flow_control.max_messages > 0 or flow_control.max_bytes > 0, flow control will be enforced at the server side (in addition to the client side).

This behavior is enabled by default and users who would like to opt-out of this feature --in case they encounter issues with server side flow control-- can pass in use_legacy_flow_control=True in SubscriberClient.subscribe().

### DIFF
--- a/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
+++ b/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
@@ -105,6 +105,9 @@ class StreamingPullManager(object):
             ``projects/{project}/subscriptions/{subscription}``.
         flow_control (~google.cloud.pubsub_v1.types.FlowControl): The flow
             control settings.
+        use_legacy_flow_control (bool): Disables enforcing flow control settings
+            at the Cloud PubSub server and uses the less accurate method of only
+            enforcing flow control at the client side.
         scheduler (~google.cloud.pubsub_v1.scheduler.Scheduler): The scheduler
             to use to process messages. If not provided, a thread pool-based
             scheduler will be used.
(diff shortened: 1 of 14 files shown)


google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -105,6 +105,9 @@ class StreamingPullManager(object):
             ``projects/{project}/subscriptions/{subscription}``.
         flow_control (~google.cloud.pubsub_v1.types.FlowControl): The flow
             control settings.
+        use_legacy_flow_control (bool): Disables enforcing flow control settings
```
pradn, line 108: I'd prefer disable_server_side_flow_control so it's right in the name. We can mention that it's legacy in the comment.

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -587,8 +596,12 @@ def _get_initial_request(self, stream_ack_deadline_seconds):
             stream_ack_deadline_seconds=stream_ack_deadline_seconds,
             subscription=self._subscription,
             client_id=self._client_id,
-            max_outstanding_messages=self._flow_control.max_messages,
-            max_outstanding_bytes=self._flow_control.max_bytes,
+            max_outstanding_messages=(
+                0 if self._use_legacy_flow_control else self._flow_control.max_messages
```
pradn, line 600: I don't quite understand this. Are we disabling client-side flow control altogether if _use_legacy_flow_control is true?

pradn (changes requested):

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
fayssalmartanigcp: No, the default right now (and with this change) is to enforce flow control as the Cloud Pub/Sub server side and also at the client side (see `maybe_pause_consumer()`, `maybe_resume_consumer()` and `_maybe_release_messages()`).

When use_legacy_flow_control=True, only the client side flow control continues to be enforced, the Cloud Pub/Sub server will not enforce any flow control. (this is the state of the world today, we are enabling server side flow control --along with the current client side-- and want to give users a path to disable this functionality if it causes issues.)

pradn: NVM, I had thought this was in the constructor, not when sending the request. This makes sense.

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
pradn: Offline convo: we're using this name cross-platform and will try to deprecate it asap.

pradn (approved):

PR #236, opened 2020-11-11
feat: Add dead lettering max delivery attempts argument

### LINKED ISSUE #235: Add optional argument for max delivery attempts when creating or updating subscription with dead lettering
To make the sample python code resemble the UI and CPS behavior a bit more we could add an optional max_delivery_attempts argument instead of setting the value to 10(when we create subscription) or 20(when we update subscription). If the user decides not to set the max_delivery_attempts argument we can set it to the default of 5 just like CPS does.

danavaziri-ga:
Add optional max_delivery_attempts argument to sample code to more closely resemble Cloud Pub/Sub behavior.

Fixes #235  🦕

### DIFF
--- a/samples/snippets/subscriber.py
+++ b/samples/snippets/subscriber.py
@@ -91,7 +91,7 @@ def create_subscription(project_id, topic_id, subscription_id):
 
 def create_subscription_with_dead_letter_topic(
     project_id, topic_id, subscription_id, dead_letter_topic_id,
-    max_delivery_attempts
+    max_delivery_attempts=5
 ):
     """Create a subscription with dead letter policy."""
     # [START pubsub_dead_letter_create_subscription]
@@ -109,7 +109,7 @@ def create_subscription_with_dead_letter_topic(
     # TODO(developer): This is an existing dead letter topic that the subscription
     # with dead letter policy will forward dead letter messages to.
     # dead_letter_topic_id = "your-dead-letter-topic-id"
-    # TODO(developer): This is an the maximum number of delivery attempts allowed 
+    # TODO(developer): This is an the maximum number of delivery attempts allowed
     # before a message gets dead lettered.
     # max_delivery_attempts = "your-max-delivery-attempts"
 
@@ -265,7 +265,7 @@ def update_push_subscription(project_id, topic_id, subscription_id, endpoint):
 
 def update_subscription_with_dead_letter_policy(
     project_id, topic_id, subscription_id, dead_letter_topic_id,
-    max_delivery_attempts
+    max_delivery_attempts=5
 ):
     """Update a subscription's dead letter policy."""
     # [START pubsub_dead_letter_update_subscription]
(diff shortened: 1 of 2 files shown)


samples/snippets/subscriber_test.py
```diff
@@ -208,7 +208,7 @@ def test_create_subscription_with_dead_letter_policy(
         pass
 
     subscriber.create_subscription_with_dead_letter_topic(
-        PROJECT_ID, TOPIC, SUBSCRIPTION_DLQ, DEAD_LETTER_TOPIC
+        PROJECT_ID, TOPIC, SUBSCRIPTION_DLQ, DEAD_LETTER_TOPIC, 10
```
leahecole, line 211: Question for this line and line 222
- What are the values 10 and 20? Why did we choose them to be 10 and 20 and not whatever the default is?
- If we do need to pass them and not use the default, please use a constant (or two if you want to have the separate values 10&20) like the other arguments passed in the tests - it makes for easier maintenance

leahecole (changes requested):

samples/snippets/subscriber_test.py
danavaziri-ga: I addressed this in a seperate commit https://github.com/googleapis/python-pubsub/pull/236/commits/7ad3b6a453cbc7f5122ffd21d6ed8d808b068ac9.

I added two constants one for the default value and one for the updated one. We could call create subscription with the default value, which is what my new commit does. However, the update subscription call should be called with a new value because we are testing to see if max_delivery_attempts changes after the call.

samples/snippets/subscriber.py
```diff
@@ -108,6 +109,9 @@ def create_subscription_with_dead_letter_topic(
     # TODO(developer): This is an existing dead letter topic that the subscription
     # with dead letter policy will forward dead letter messages to.
     # dead_letter_topic_id = "your-dead-letter-topic-id"
+    # TODO(developer): This is an the maximum number of delivery attempts allowed
```
anguillanneuf, line 112: "This is the maximum number of delivery attempts allowed for a message before it gets delivered to a dead letter topic."

samples/snippets/subscriber.py
```diff
@@ -108,6 +109,9 @@ def create_subscription_with_dead_letter_topic(
...
+    # before a message gets dead lettered.
+    # max_delivery_attempts = "your-max-delivery-attempts"
```
anguillanneuf, line 114: Use an integer as an example. `max_delivery_attempts = 5`

samples/snippets/subscriber.py
```diff
@@ -276,6 +282,9 @@ def update_subscription_with_dead_letter_policy(
...
```
anguillanneuf, line 287: Ditto.

anguillanneuf (commented):
A few nits then LGTM.

samples/snippets/subscriber.py
anguillanneuf: nit: I think you still forgot the cross out "an" in "This is an the"

danavaziri-ga: Thank you for pointing that out!

anguillanneuf (approved):

leahecole (approved):

Issue #222, opened 2020-10-13
Allow retry and timeout settings on the publisher client


anguillanneuf:
Allow retry and timeout settings on the [Publisher](https://googleapis.dev/python/pubsub/latest/publisher/api/client.html). 

`timeout=` should take `google.api_core.timeout.ExponentialTimeout` and `google.api_core.timeout.ConstantTimeout` (also `LogicalTimeout` in future work).

The goal is be able to set custom retry and custom timeout on the PublisherClient, like this:

```python
custom_retry = api_core.retry.Retry(
    initial=0.250,  # seconds (default: 0.1)
    maximum=90.0,  # seconds (default: 60.0)
    multiplier=1.45,  # default: 1.3
    deadline=300.0,  # seconds (default: 60.0)
    predicate=api_core.retry.if_exception_type(
        api_core.exceptions.Aborted,
        api_core.exceptions.DeadlineExceeded,
        api_core.exceptions.InternalServerError,
        api_core.exceptions.ResourceExhausted,
        api_core.exceptions.ServiceUnavailable,
        api_core.exceptions.Unknown,
        api_core.exceptions.Cancelled,
    ),
)

custom_timeout=api_core.timeout.ExponentialTimeout(
    initial=1.0,  
    maximum=10.0,  
    multiplier=1.0,  
    deadline=300.0,  
)

# Something like this:
publisher = pubsub_v1.PublisherClient(
    publisher_options = pubsub_v1.types.PublisherOptions(
        retry=custom_retry,
        timeout=custom_timeout,
    ),
)
```

Here are the defaults and how they are set currently: https://github.com/googleapis/googleapis/blob/3a935fab757e09c72afd4aa121147a4c97dccc3e/google/pubsub/v1/pubsub_grpc_service_config.json#L42-L62

Remember to update `UPGRADING.md` as well.

cguardia:
@anguillanneuf took a bit of time to really get started here, but I'm on it now. How about other api calls that take a float timeout, like  `get_topic` or `list_topics`? Should those also take the api_core timeout objects?

anguillanneuf:
@cguardia Those can be left alone because the default settings in `pubsub_grpc_service_config.json` are good as they are.

Adding @kamalaboulhosn for comments too.

kamalaboulhosn:
[...]

PR #239, opened 2020-11-16
wip: Allow retry and timeout settings on the publisher client

cguardia:
This is a feature requested on issue #222. There are a couple of issues to discuss regarding the implementation.

### DIFF
--- a/google/cloud/pubsub_v1/publisher/client.py
+++ b/google/cloud/pubsub_v1/publisher/client.py
@@ -272,7 +272,13 @@ class Client(object):
                 sequencer.unpause()
 
     def publish(
-        self, topic, data, ordering_key="", retry=gapic_v1.method.DEFAULT, **attrs
+        self,
+        topic,
+        data,
+        ordering_key="",
+        retry=gapic_v1.method.DEFAULT,
+        timeout=gapic_v1.method.DEFAULT,
+        **attrs
     ):
         """Publish a single message.
 
--- a/google/pubsub_v1/services/publisher/client.py
+++ b/google/pubsub_v1/services/publisher/client.py
@@ -26,6 +26,7 @@ from google.api_core import client_options as client_options_lib  # type: ignore
 from google.api_core import exceptions  # type: ignore
 from google.api_core import gapic_v1  # type: ignore
 from google.api_core import retry as retries  # type: ignore
+from google.api_core import timeout as timeouts  # type: ignore
 from google.auth import credentials  # type: ignore
 from google.auth.transport import mtls  # type: ignore
 from google.auth.transport.grpc import SslCredentials  # type: ignore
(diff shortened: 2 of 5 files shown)


google/pubsub_v1/services/publisher/client.py
```diff
@@ -26,6 +26,7 @@
 from google.api_core import exceptions  # type: ignore
 from google.api_core import gapic_v1  # type: ignore
 from google.api_core import retry as retries  # type: ignore
+from google.api_core import timeout as timeouts  # type: ignore
```
cguardia, line 29: This is a generated file, but I included it in this PR so that it can be seen what the synth changes will do to it.

cguardia:
@anguillanneuf coming back to this after a few days absence. I'm not sure if this is the best way to do this. I added the publisher options and updated the docs and client, but to change the actual method definitions, I had to modify the synth code. What is the recommended way to make this kind of change?

cguardia:
@pradn Hi, for issue #222 we decided to have custom retry and timeout settings for the publisher options, and also allow the user to send them on all API calls. This is my first experience with auto-generated code, so I'm not sure I'm doing this right. Please take a look when you can and comment.

anguillanneuf:
Thanks @cguardia for the PR. Setting timeout on the publisher client using publish settings looks right to me. @pradn thoughts?

google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -272,7 +272,13 @@ def resume_publish(self, topic, ordering_key):
                 sequencer.unpause()
 
     def publish(
-        self, topic, data, ordering_key="", retry=gapic_v1.method.DEFAULT, **attrs
+        self,
+        topic,
+        data,
+        ordering_key="",
+        retry=gapic_v1.method.DEFAULT,
```
pradn, line 279: If retry and timeout are added to PublisherOptions, then there's no need to add these parameters here, yes?

pradn:
The approach of adding timeout to the PublisherOptions looks good to me. You'll have to plumb the timeout down through the sequencers to where the RPC is made.

meredithslota:
What is the status of this PR? It feels like it's lingering but I'm not sure if it's still a WIP or if it's ready to go. Thanks!

plamut:
@meredithslota I queried for its status. If nothing else, I can probably take over, as I expect to find some free capacity in the near future.

google/cloud/pubsub_v1/publisher/client.py
plamut: Somebody might want to override these default options for a few particular `publish()` calls?

I'd say we keep these two arguments for flexibility, as well as for consistency with other Publisher methods. Besides, the `publish()` method signature already includes the `retry` parameter and thus shouldn't be removed, and it makes more sense to retain `timeout` with it.

plamut:
Closing in favor of #299.

PR #266, opened 2020-12-23
fix: Modify synth.py to update grpc transport options

dpcollins-google:
Includes manual changes and synth.py regeneration.


busunkim96:
I did a rebase on master, tweaked the regex, and ran synthtool. The diff has some unrelated file changes because there are template changes and a new proto.

busunkim96 (approved):

PR #269, opened 2020-12-29
fix: Don't open the google.cloud package by adding pubsub.py

dpcollins-google:
Instead, create a subpackage, google.cloud.pubsub and add an `__init__.py` file to it.

This should have no user-facing effect.


dpcollins-google:
Added tianzi and bu sun to confirm that this should be a no-op to users.

busunkim96 (approved):

PR #275, opened 2021-01-16
fix: client version missing from the user agent header

### LINKED ISSUE #200: Auto generated code get_distribution method through `DistributionNotFound` error
Try statement never executed because of `get_distribution(google-pubsub)`, it should be get_distribution("google-cloud-pubsub",)
https://github.com/googleapis/python-pubsub/blob/6b9eec81ccee81ab93646eaf7652139fc218ed36/google/pubsub_v1/services/publisher/transports/base.py#L34-L37

also found same thing at various places:
https://github.com/googleapis/python-pubsub/blob/6b9eec81ccee81ab93646eaf7652139fc218ed36/google/pubsub_v1/services/subscriber/transports/base.py#L36

https://github.com/googleapis/python-pubsub/blob/6b9eec81ccee81ab93646eaf7652139fc218ed36/google/pubsub_v1/services/publisher/async_client.py#L1093

https://github.com/googleapis/python-pubsub/blob/6b9eec81ccee81ab93646eaf7652139fc218ed36/google/pubsub_v1/services/publisher/client.py#L1199

https://github.com/googleapis/python-pubsub/blob/6b9eec81ccee81ab93646eaf7652139fc218ed36/google/pubsub_v1/services/subscriber/client.py#L1904

plamut:
Fixes #200.

This PR patches the generated code so that client library version is actually included in the `x-goog-api-client` metrics key.

PR checlist:

### DIFF
--- a/synth.py
+++ b/synth.py
@@ -97,6 +97,23 @@ s.replace(
         \g<0>""",
 )
 
+# Make sure that client library version is present in user agent header.
+s.replace(
+    [
+        "google/pubsub_v1/services/publisher/async_client.py",
+        "google/pubsub_v1/services/publisher/client.py",
+        "google/pubsub_v1/services/publisher/transports/base.py",
+        "google/pubsub_v1/services/schema_service/async_client.py",
+        "google/pubsub_v1/services/schema_service/client.py",
+        "google/pubsub_v1/services/schema_service/transports/base.py",
+        "google/pubsub_v1/services/subscriber/async_client.py",
+        "google/pubsub_v1/services/subscriber/client.py",
+        "google/pubsub_v1/services/subscriber/transports/base.py",
+    ],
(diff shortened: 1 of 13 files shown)


synth.py
```diff
@@ -97,6 +97,23 @@
         \g<0>""",
 )
 
+# Make sure that client library version is present in user agent header.
+s.replace(
```
tswast, line 101: Is this something we can update in the Bazel config to avoid manual overrides? https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/BUILD.bazel

tswast (approved):

synth.py
anguillanneuf: Not quite familiar with BUILD.bazel, but it looks like pubsub should belong in https://github.com/googleapis/googleapis/tree/master/google/cloud just as pubsublite?

PR #292, opened 2021-02-12
feat: add graceful streaming pull shutdown

### LINKED ISSUE #276: Gracefully shutdown subscriber
Hey,

#### Environment details

  - OS type and version: Ubuntu 18.04
  - Python version: 3.6.9
  - pip version: 9.0.1
  - `google-cloud-pubsub` version: 2.2.0

#### Code example

Sample code for a subscriber from Google's guide with minor changes that will show my issue:
```python

from concurrent.futures import TimeoutError
from time import sleep

# TODO(developer)
# project_id = "your-project-id"
# subscription_id = "your-subscription-id"
# Number of seconds the subscriber should listen for messages
# timeout = 5.0

# The `subscription_path` method creates a fully qualified identifier
# in the form `projects/{project_id}/subscriptions/{subscription_id}`
subscription_path = subscriber.subscription_path(project_id, subscription_id)

def callback(message):
    print(f"Received {message}.")
    sleep(30)
    message.ack()
    print("Acked")

streaming_pull_future = subscriber.subscribe(subscription_path, callback=callback)
print(f"Listening for messages on {subscription_path}..\n")

# Wrap subscriber in a 'with' block to automatically call close() when done.
with subscriber:
    sleep(10)
    streaming_pull_future.result()
```
From https://cloud.google.com/pubsub/docs/pull

I expect this code to stop pulling messages and finish the running messages and then exits.

Actually this code stops pulling messages and finish executing the running messages but it does not ack the messages. The .ack() happens but the server does not receive the ack, so next run the same messages return again.

1. Why doesn't the server receives the ack?

2. How to gracefully shutdown the subscriber?

3. What is the expected behavior of .cancel()?

Corresponding stackoverflow question:
https://stackoverflow.com/questions/65761254/pub-sub-python-client-gracefully-shutdown-subscriber

Thank you!

plamut:
Closes #276.

This PR implements an optional blocking streaming pull shutdown. "Blocking" means  blocking on `streaming_future.cancel()` until all currently executing user callbacks complete, as well as dispatching any message ACK or NACK requests these callbacks produce.

The PR also makes sure that any received messages sitting in internal queues - but not yet dispatched to the callbacks - are automatically NACK-ed on shutdown to speed up their re-delivery (the backend is not left waiting for their ACK deadlines to expire).

PR checklist:

### DIFF
--- a/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
+++ b/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
@@ -99,9 +99,6 @@ class Dispatcher(object):
             ValueError: If ``action`` isn't one of the expected actions
                 "ack", "drop", "lease", "modify_ack_deadline" or "nack".
         """
-        if not self._manager.is_active:
-            return
-
         batched_commands = collections.defaultdict(list)
 
         for item in items:
--- a/google/cloud/pubsub_v1/subscriber/_protocol/heartbeater.py
+++ b/google/cloud/pubsub_v1/subscriber/_protocol/heartbeater.py
@@ -35,10 +35,11 @@ class Heartbeater(object):
         self._period = period
 
     def heartbeat(self):
-        """Periodically send heartbeats."""
-        while self._manager.is_active and not self._stop_event.is_set():
-            self._manager.heartbeat()
-            _LOGGER.debug("Sent heartbeat.")
+        """Periodically send streaming pull heartbeats.
+        """
+        while not self._stop_event.is_set():
+            if self._manager.heartbeat():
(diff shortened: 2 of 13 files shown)


plamut:
It appears that ACK requests sent after the shutdown somehow get lost, even though the dispatcher processes them and sends them through the streaming pull manager's (`send()` over unary RPC). I also reproduced the same on the `master` branch... weird. :confused:

kamalaboulhosn (approved):

anguillanneuf:
This trickled down to `python-pubsublite` and caused https://github.com/googleapis/python-pubsublite/issues/95.

PR #299, opened 2021-02-26
feat: support customizable retry and timeout settings on the publisher client

### LINKED ISSUE #222: Allow retry and timeout settings on the publisher client

plamut:
Closes #222.
Supersedes #239.

This PR adds support for API core timeout classes to the publisher class, replacing the previous `float` timeouts.

**PR checklist:**


cguardia:
@googlebot I consent.

jimfulton (commented):
I think the type UNION should admit floats and ints.

I also think the union should be defined once and reused.

There's an obscene amount of repetition in this package.  It was already there, but may be can not add to it. :)

google/cloud/pubsub_v1/types.py
```diff
@@ -37,6 +39,16 @@
 from google.pubsub_v1.types import pubsub as pubsub_gapic_types
 
 
+TimeoutType = Union[
+    None,
```
jimfulton: Did we mean to add `None`?  Was this my fault?

This feels like a breaking change.  We didn't accept `None` before.

google/pubsub_v1/services/publisher/async_client.py
```diff
@@ -315,6 +313,9 @@ async def update_topic(
             ),
         )
 
+        if timeout is None or isinstance(timeout, (int, float)):
+            timeout = timeouts.ConstantTimeout(timeout)
```
jimfulton, line 317: We shouldn't need except for the `None` case, which is new.  I don't think we should accept `None`.  Apologies if `None` came from me.

`int` and `float` are handled here: 

https://github.com/googleapis/python-api-core/blob/7337c6b123735fb9ae5d0e54b4399275719c020c/google/api_core/gapic_v1/method.py#L66-L67

The same comment applies to other places where you added this.

jimfulton: And, of course, not DRY.

google/cloud/pubsub_v1/publisher/_sequencer/base.py
```diff
@@ -45,18 +47,17 @@ def unpause(self, message):  # pragma: NO COVER
 
     @staticmethod
     @abc.abstractmethod
-    def publish(self, message, retry=None, timeout=None):  # pragma: NO COVER
+    def publish(
+        self, message, retry=None, timeout: gapic_types.TimeoutType = None
```
jimfulton, line 51: Can we use `DEFAULT` rather than `None`?  The implementation defaults to `DEFAULT`.

jimfulton (commented):
Thanks for these changes.

My only complaint is that we added None as a valid timeout, which causes some pain.

plamut:
@jimfulton  I added none, because I noticed that [ConstantTimeout](https://github.com/googleapis/python-api-core/blob/7337c6b123735fb9ae5d0e54b4399275719c020c/google/api_core/timeout.py#L71-L82) actually supports it.

plamut:
Flaky samples tests, 500 internal.

google/cloud/pubsub_v1/publisher/_sequencer/base.py
plamut: That makes sense, the ABC should be consistent.

jimfulton:
> @jimfulton I added none, because I noticed that [ConstantTimeout](https://github.com/googleapis/python-api-core/blob/7337c6b123735fb9ae5d0e54b4399275719c020c/google/api_core/timeout.py#L71-L82) actually supports it.
> 
> (not saying that explicitly disabling any timeouts is a good idea, but we already support that automaticlally if we allow instances of `ConstantTimeout`)

That's fair, but at least you could argue that we should make people work to disable timeouts.  


BTW, I tried to figure out what the effect of passing `None` to `ConstantTimout` would be and got lost in the web of indirection.  :)  To convince myself that I knew what was going on I'd have to spend some quality time with `pdb`, and I don't have time for that today.

In any case, if we want to accept `None`, I think we should handle that in one place (https://github.com/googleapis/python-api-core/blob/7337c6b123735fb9ae5d0e54b4399275719c020c/google/api_core/gapic_v1/method.py#L66-L67) so we don't repeat the logic in many places.

plamut:
I don't see a good reason why somebody wanted to potentially block indefinitely, thus I'm fine with removing support for a plain `None` (one can always pass in `ConstantTimeout(None)`, if they _really_ want to). We can then also get rid of the "convert constant to object" code in the generated code, because number constants are already handled in `api-core`.

plamut:
> BTW, I tried to figure out what the effect of passing `None` to `ConstantTimout` would be and got lost in the web of indirection. :) 

If I read it correctly, a `ConstantTimeout(None)` is used as a RPC method [decorator](https://github.com/googleapis/python-api-core/blob/7337c6b123735fb9ae5d0e54b4399275719c020c/google/api_core/gapic_v1/method.py#L132). What `ConstantTimeout` does is just [setting](https://github.com/googleapis/python-api-core/blob/7337c6b123735fb9ae5d0e54b4399275719c020c/google/api_core/timeout.py#L98-L102) the `timeout` argument to the wrapped method to the given value (`None` in this case).

jimfulton (approved):

plamut:
Note to self - this PR modifies `synth.py`, thus after merging, we also need to update https://github.com/googleapis/python-pubsub/pull/411

cc: @parthea

Issue #264, opened 2020-12-21
samples: add type hints to all samples 


anguillanneuf:
Add type hints so setting "enforce_type_hints" to `True` won't cause lint failures.

https://github.com/googleapis/python-pubsub/blob/258414727827e56d2678ada28b9bbd88b5c6aa6f/samples/snippets/noxfile_config.py#L26-L28

plamut:
@anguillanneuf What's the status of this, have you started working on type hints yet?

If not, I can probably handle this, I'll have some spare cycles.

anguillanneuf:
@plamut Thanks for the ping. If you can handle it, I will review. Thanks again!

PR #306, opened 2021-03-03
samples: add schema samples

anguillanneuf:
- [ ] create avro schema
- [ ] create proto schema
- [ ] get schema
- [ ] list schemas
- [ ] delete scheam
- [ ] create topic with schema
- [ ] publish avro records
- [ ] publish proto messages
- [ ] subscribe with avro schema
- [ ] subscribe with proto schema


samples/snippets/README.rst
```diff
@@ -266,6 +266,48 @@ To run this sample:
 
 
 
+Schema
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+.. image:: https://gstatic.com/cloudssh/images/open-btn.png
+   :target: https://console.cloud.google.com/cloudshell/open?git_repo=https://github.com//googleapis/python-pubsub&page=editor&open_in_editor=samples/snippets/iam.py,samples/snippets/README.rst
```
leahecole, line 273:
```suggestion
   :target: https://console.cloud.google.com/cloudshell/open?git_repo=https://github.com/googleapis/python-pubsub&page=editor&open_in_editor=samples/snippets/iam.py,samples/snippets/README.rst
```

leahecole: As is, the button triggers a warning that say "this repo isn't maintained by google, do you trust it, blah blah" and then when I said trust it went to the last repo I'd cloned in cloud shell, which was from another button

Additionally, it looks like this button opens up the `README` and `snippets/iam` -is that intended or is it a copy paste error?

leahecole (dismissed):
I did a quicker review for now but have two questions @anguillanneuf - 
1) Is it possible to have the sample not be run using argparse w/ command line args? This is something I admit we do not do a good job of documenting and I need to update the authoring guide to mention. It's a lot harder to maintain and test the various combinations/permutations of the command-line arguments, so we try to avoid it where we can. 
2) Given that this is a new sample, could you please add in [type hints](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/AUTHORING_GUIDE.md#documenting-types)? You can add [a line](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/16941a107eec9e1437a7a1422e5139701a32a137/run/idp-sql/noxfile_config.py#L27) to a `noxfile_config.py` file that will make the linter look for type hints. 

feel free to set up time if you want to talk about this especially my pushback wrt argparse

samples/snippets/schema.py
```diff
@@ -189,24 +191,119 @@ def create_topic_with_schema(project_id, topic_id, schema_id, encoding):
...
+            encoder = BinaryEncoder(bout)
+            writer.write(record, encoder)
+            data = bout.getvalue()
+
+            print("Preparing a binary-encoded message.")
+        elif encoding == Encoding.JSON:
+            data = json.dumps(record).encode("utf-8")
+            print("Preparing a JSON-encoded message.")
+        else:
+            raise InvalidArgument(ValueError)
```
plamut, line 236: (applies to multiple places)

The first argument to `InvalidArgument` should be a string (error messages). Also, `InvalidArgument` inherits from `GoogleAPICallError`, which does not seem like a correct exception type, as it's not a result of an API call.

BTW, the only way to hit this branch is if topic's schema settings encoding is `ENCODING_UNSPECIFIED`, but that's the info retrieved from the server? (i.e. not something we would pass as an argument)

samples/snippets/schema.py
```diff
@@ -0,0 +1,489 @@
...
+
+    schema_client = SchemaServiceClient()
+    schema_path = schema_client.schema_path(project_id, schema_id)
+
+    if message_encoding == "BINARY":
+        encoding = Encoding.BINARY
+    elif message_encoding == "JSON":
+        encoding = Encoding.JSON
+    else:
+        encoding = Encoding.ENCODING_UNSPECIFIED
```
plamut, line 178: Any reason to not use the `types.Encoding` enum directly?

If we use string constants in the sample, a good portion of users will follow suit, but that requires some extra boilerplate code and is also susceptible to typos.

plamut (changes requested):
Looks good in general, but I did found a few things that we might want to have another look at.

samples/snippets/README.rst
anguillanneuf: >  it went to the last repo I'd cloned in cloud shell
Not bad actually. It goes to the current repo for me despite having other repos in the root directory. 

Copy-paste error corrected.

samples/snippets/schema.py
```diff
@@ -0,0 +1,298 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""This application demonstrates how to perform basic schema operations
+using the Cloud Pub/Sub API.
+
+For more information, see the README.md under /pubsub and the documentation
+at https://cloud.google.com/pubsub/docs.
```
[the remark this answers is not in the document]
anguillanneuf: A better link is https://cloud.google.com/pubsub/docs/schemas. Updated.

samples/snippets/schema.py
```diff
@@ -0,0 +1,298 @@
...
+"""
+
+import argparse
+
+
+def create_avro_schema(project_id, schema_id, avsc_file):
+    """Create a schema resource from an Avro schema file formatted in JSON."""
```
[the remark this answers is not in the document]
anguillanneuf: AFAIK Avro schemas are formatted in JSON. A little redundant but just want to point it out anyway.

samples/snippets/schema_test.py
```diff
@@ -0,0 +1,204 @@
...
+
+from google.api_core.exceptions import NotFound
+from google.cloud import pubsub_v1
+import pytest
+
+import schema
+
+UUID = uuid.uuid4().hex
+PY_VERSION = f"{sys.version_info.major}.{sys.version_info.minor}"
+PROJECT_ID = os.environ["GOOGLE_CLOUD_PROJECT"]
```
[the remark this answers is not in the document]
anguillanneuf: Done.

samples/snippets/schema.py
anguillanneuf: I rewrote this part to exit(0). Is that better?

samples/snippets/schema.py
anguillanneuf: The main reason I'm writing this way is because I'm not sure how to pass `types.Encoding.BINARY` in place of `BINARY` in the following command: 

`python schema.py $PROJECT_ID publish-avro $TOPIC_ID resources/us-states.avsc BINARY`

anguillanneuf:
@leahecole 
1. Discussed offline.
2. See #264 

@plamut 
Addressed all of them but left two convo unresolved for you to take another look.

samples/snippets/schema.py
plamut: The approach seems reasonable IMO. :+1:

samples/snippets/schema.py
plamut: Ah, this can come through cmd arguments, fine then. I like the improvement that restricts the valid argument choices to "BINARY" or "JSON".

plamut (approved):
LGTM.

I noticed that the samples shell that opens uses Python 2.7, which results in a syntax error where f-strings are used, but I suppose that's an external thing... or is there a way to explicitly require that Python 3.6+ is used with the shell?

anguillanneuf:
@plamut I just tried with a new account. I can confirm that Python 3.7 comes pre-installed in Cloud Shell. You need to run it with `python3` instead of `python`.

PR #315, opened 2021-03-10
revert: remove graceful streaming pull shutdown

anguillanneuf:
Reverts googleapis/python-pubsub#292

Release 2.4.0 which contains this feature has been yanked from PyPi: https://pypi.org/project/google-cloud-pubsub/

Graceful streaming pull shutdown introduces an optional wait on callbacks on streaming pull futures. It would otherwise break the streaming pull future returned by the Pub/Sub Lite subscriber client. See https://github.com/googleapis/python-pubsublite/issues/95

### DIFF
--- a/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
+++ b/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
@@ -99,6 +99,9 @@ class Dispatcher(object):
             ValueError: If ``action`` isn't one of the expected actions
                 "ack", "drop", "lease", "modify_ack_deadline" or "nack".
         """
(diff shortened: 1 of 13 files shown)


busunkim96:
@anguillanneuf Sorry to edit the title on you again. 
 
I checked the conventional commit spec and it turns out there's a `revert:` type. 😄 
 https://www.conventionalcommits.org/en/v1.0.0/#how-does-conventional-commits-handle-revert-commits aaand release-please knows what to do with it. https://github.com/googleapis/release-please/blob/d8ac96721de299f89fb435dc1650bf5d7dce8857/src/releasers/python.ts#L35

PR #319, opened 2021-03-10
Revert "revert: remove graceful streaming pull shutdown"

### LINKED ISSUE #318: Graceful streaming pull shutdown breaks PubSub Lite
#292 added an option to gracefully wait for any active message callbacks on streaming pull shutdown, but the addition of the new optional parameter `await_msg_callbacks` broke cancelling the PubSub Lite ([issue](https://github.com/googleapis/python-pubsublite/issues/95)).

As discussed offline, this parameter should be removed from the streaming pull future, and instead implemented as a subscriber option.

plamut:
Reverts googleapis/python-pubsub#315 that reverted #292.
Fixes #318.

A full revert re-introduces several issues fixed in #292. This PR reverts the revert, but also makes the (less radical) changes that avoid the error in PubSub Lite client.


plamut:
Superseded by #320.

PR #320, opened 2021-03-10
fix: move await_msg_callbacks flag to subscribe() method

### LINKED ISSUE #318: Graceful streaming pull shutdown breaks PubSub Lite

plamut:
Reverts googleapis/python-pubsub#315 that reverted #292.
Fixes #318.
Supersedes #319.

A full revert of #292 re-introduces several issues that have been fixed with it. This PR thus reverts the revert, but also makes the necessary (less radical) changes that avoid the error in PubSub Lite client - the `StreamingPullFuture` is now intact, and the `await_msg_callbacks` parameter of the `cancel()` method has been moved to subscriber client's `subscribe()` method.

Please test the PubSub Lite client with this PR branch to verify that it indeed doesn't break it anymore, thanks!

**PR checklist:**

### DIFF
--- a/google/cloud/pubsub_v1/subscriber/client.py
+++ b/google/cloud/pubsub_v1/subscriber/client.py
@@ -201,11 +201,13 @@ class Client(object):
                 how callbacks are executed concurrently. This object must not be shared
                 across multiple SubscriberClients.
             await_callbacks_on_shutdown (bool):
-                If ``True``, the returned future's ``close()`` method will block until
-                the background stream and its helper threads have has been terminated,
-                and all currently executing message callbacks are done processing.
+                If ``True``, after canceling the returned future, the latter's
+                ``result()`` method will block until the background stream and its
+                helper threads have been terminated, and all currently executing message
+                callbacks are done processing.
 
-                If ``False`` (default), the returned future's ``close()`` method returns
+                If ``False`` (default), the returned future's ``result()`` method will
+                not block after canceling the future. The method will instead return
                 immediately after the background stream and its helper threads have been
                 terminated, but some of the message callback threads might still be
                 running at that point.
--- a/google/cloud/pubsub_v1/subscriber/client.py
+++ b/google/cloud/pubsub_v1/subscriber/client.py
@@ -122,6 +122,7 @@ class Client(object):
         flow_control=(),
         scheduler=None,
         use_legacy_flow_control=False,
+        await_callbacks_on_shutdown=False,
     ):
         """Asynchronously start receiving messages on a given subscription.
 
@@ -199,6 +200,15 @@ class Client(object):
                 *scheduler* to use when executing the callback. This controls
                 how callbacks are executed concurrently. This object must not be shared
                 across multiple SubscriberClients.
+            await_callbacks_on_shutdown (bool):
+                If ``True``, the returned future's ``close()`` method will block until
+                the background stream and its helper threads have has been terminated,
+                and all currently executing message callbacks are done processing.
+
+                If ``False`` (default), the returned future's ``close()`` method returns
+                immediately after the background stream and its helper threads have been
+                terminated, but some of the message callback threads might still be
+                running at that point.
 
         Returns:
             A :class:`~google.cloud.pubsub_v1.subscriber.futures.StreamingPullFuture`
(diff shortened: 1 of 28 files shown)


dpcollins-google:
As far as I can tell, this will not break Pub/Sub Lite python. I will separately work on making it clear which API surfaces Pub/Sub Lite overrides so it is at least clear in the future what kind of changes would break the other client, and figuring out how to test this.

plamut:
FWIW, as per offline discussion - the `cancel()` itself should not block for "long", the following was proposed:
```python
try:
   streaming_future.result()
except Something:
    streaming_future.cancel()  # should "fast"
    streaming_future.result()  # this should block until msg callbacks complete, if await_msg_callbacks was set
 ```

**Review by anguillanneuf — APPROVED** [review]
Thanks Peter for the quick fix. I gave it a try and it did fix https://github.com/googleapis/python-pubsublite/issues/95.

LGTM.

**Comment by plamut** [comment]
Good to hear that it works.

@dpcollins-google made a good point that the streaming pull future's `cancel()` should not block, but instead any blocking operations should be done when `.result()` is called.

The `future.cancel()` has actually been a blocking call since forever, I think the behavior dates back to Ancient Rome when threads have not been invented yet. The call blocks until the shutdown is complete, which is usually quite fast, but a graceful shutdown makes it much more noticeable.

The solution is to move the shutdown operations into their own thread, similarly to what we [already do](https://github.com/googleapis/python-pubsub/blob/4dc89fc0fa1c4495bd0d06ee17596390dcd8a497/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py#L723-L729) when we close the manager due to a terminal stream error. I'm probably offline tomorrow, but can add this before the end of the week.

If we are not in a hurry with the release, I'd say we wait for this improvement before merging. Otherwise just remove the label and I'll submit a separate PR.

**Comment by plamut** [comment]
Made the `streaming_pull_future.cancel()` call non-blocking, the user code should block when calling `future.result()` after canceling it.

I will follow up with a sample, but if the PR is merged in the meantime, I can open a separate PR.

**Comment by plamut** [comment]
cla/google status update is stuck.

**Comment by anguillanneuf** [comment]
I force rescanned to get the CLA signed.

File: google/cloud/pubsub_v1/subscriber/client.py [hunk 1]
```diff
@@ -199,6 +200,15 @@ def callback(message):
                 *scheduler* to use when executing the callback. This controls
                 how callbacks are executed concurrently. This object must not be shared
                 across multiple SubscriberClients.
+            await_callbacks_on_shutdown (bool):
+                If ``True``, the returned future's ``close()`` method will block until
```
anguillanneuf, line 204:
```suggestion
                If ``True``, the returned future's ``cancel()`` method will block until
```

google/cloud/pubsub_v1/subscriber/client.py
```diff
@@ -199,6 +200,15 @@ def callback(message):
...
+                the background stream and its helper threads have has been terminated,
+                and all currently executing message callbacks are done processing.
+
+                If ``False`` (default), the returned future's ``close()`` method returns
```
anguillanneuf, line 208:
```suggestion
                If ``False`` (default), the returned future's ``cancel()`` method returns
```

google/cloud/pubsub_v1/subscriber/client.py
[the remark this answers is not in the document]
plamut: Ah, that's an incorrectly adjusted copy pasta from the `StreamingPullManager`, nice catch.

The future has a `cancel()` method, yes, but it was specifically made to _not_ block - it's the `result()` method that can optional block after the future has been canceled.

plamut: The future's `cancel()` method has been specifically changed to _not_ block, it's the `result()` method that can now optionally block after canceling the future - will fix the docstring.

pradn:
I'll take a look at this on Monday.

pradn:
Apologies, got caught up today. Tomorrow for sure. :)

pradn (approved):

gnagel:
Very excited to see this fix merged in, thanks for the awesome work here :-)

PR #324, opened 2021-03-11
fix: Remove EXPERIMENTAL tag for ordering keys in publisher/client.py

pradn:


plamut:
Thanks for catching and addressing that.

I'm not entirely sure why the samples tests fail, is the PR branch based on the latest `master`? BTW, I also suggest merging this with #323, so that we don't end up with a seemingly duplicate entry in release notes.

kamalaboulhosn (approved):

dinagraves (approved):

PR #361, opened 2021-04-01
chore: fix streaming pull close test flakiness

### LINKED ISSUE #352: tests.unit.pubsub_v1.subscriber.test_streaming_pull_manager: test_close_inactive_consumer failed
This test failed!

To configure my behavior, see [the Flaky Bot documentation](https://github.com/googleapis/repo-automation-bots/tree/master/packages/flakybot).

If I'm commenting on this issue too often, add the `flakybot: quiet` label and
I will stop commenting.

---

commit: d40d02713c8c189937ae5c21d099b88a3131a59f
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/edcc31f7-ba95-47d7-87a6-22509f4099c6), [Sponge](http://sponge2/edcc31f7-ba95-47d7-87a6-22509f4099c6)
status: failed
<details><summary>Test output</summary><br><pre>def test_close_inactive_consumer():
        (
            manager,
            consumer,
            dispatcher,
            leaser,
            heartbeater,
            scheduler,
        ) = make_running_manager()
        consumer.is_active = False
    
        manager.close()
    
        consumer.stop.assert_not_called()
        leaser.stop.assert_called_once()
>       dispatcher.stop.assert_called_once()

tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py:589: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

_mock_self = <MagicMock name='mock.stop' spec='function' id='139822451967824'>

    def assert_called_once(_mock_self):
        """assert that the mock was called only once.
        """
        self = _mock_self
        if not self.call_count == 1:
            msg = ("Expected '%s' to have been called once. Called %s times.%s"
                   % (self._mock_name or 'mock',
                      self.call_count,
                      self._calls_repr()))
>           raise AssertionError(msg)
E           AssertionError: Expected 'stop' to have been called once. Called 0 times.

.nox/unit-3-7/lib/python3.7/site-packages/mock/mock.py:891: AssertionError</pre></details>

### LINKED ISSUE #373: tests.unit.pubsub_v1.subscriber.test_streaming_pull_manager: test_close failed
This test failed!

To configure my behavior, see [the Flaky Bot documentation](https://github.com/googleapis/repo-automation-bots/tree/master/packages/flakybot).

If I'm commenting on this issue too often, add the `flakybot: quiet` label and
I will stop commenting.

---

commit: c707f81fdadd118bd2bc48aebe74d453f35cad0d
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/819aafcc-7191-40c3-89cb-221d25dbffa1), [Sponge](http://sponge2/819aafcc-7191-40c3-89cb-221d25dbffa1)
status: failed
<details><summary>Test output</summary><br><pre>def test_close():
        (
            manager,
            consumer,
            dispatcher,
            leaser,
            heartbeater,
            scheduler,
        ) = make_running_manager()
    
        manager.close()
    
        consumer.stop.assert_called_once()
        leaser.stop.assert_called_once()
        dispatcher.stop.assert_called_once()
>       heartbeater.stop.assert_called_once()

tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py:568: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

_mock_self = <MagicMock name='mock.stop' spec='function' id='140462164020768'>

    def assert_called_once(_mock_self):
        """assert that the mock was called only once.
        """
        self = _mock_self
        if not self.call_count == 1:
            msg = ("Expected '%s' to have been called once. Called %s times.%s"
                   % (self._mock_name or 'mock',
                      self.call_count,
                      self._calls_repr()))
>           raise AssertionError(msg)
E           AssertionError: Expected 'stop' to have been called once. Called 0 times.

.nox/unit-3-9/lib/python3.9/site-packages/mock/mock.py:891: AssertionError</pre></details>

plamut:
Fixes #352.
Fixes #373.

This PR fixes unit test flakiness by waiting for the streaming pull manager shutdown thread to finish running before making any assertions.

**PR checklist**

### DIFF
--- a/tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
+++ b/tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
@@ -550,6 +550,26 @@ def make_running_manager(**kwargs):
     )
 
 
+def await_manager_shutdown(timeout=None):  # pragma: NO COVER
+    # NOTE: This method should be called after manager.close(), i.e. after the shutdown
+    # thread has been started.
+    shutdown_thread = next(
+        (
+            thread
+            for thread in threading.enumerate()
+            if thread.name == streaming_pull_manager._REGULAR_SHUTDOWN_THREAD_NAME
+        ),
+        None,
+    )
+
+    if shutdown_thread is None:
+        return  # Shutdown already finished.
+
+    shutdown_thread.join(timeout=timeout)
+    if shutdown_thread.is_alive():
+        pytest.fail("Shutdown not completed in time.")
+
+
 def test_close():
     (
         manager,
@@ -561,6 +581,7 @@ def test_close():
     ) = make_running_manager()
 
     manager.close()
+    await_manager_shutdown(timeout=3)
 
     consumer.stop.assert_called_once()
     leaser.stop.assert_called_once()
@@ -583,6 +604,7 @@ def test_close_inactive_consumer():
     consumer.is_active = False
 
     manager.close()
+    await_manager_shutdown(timeout=3)
 
     consumer.stop.assert_not_called()
     leaser.stop.assert_called_once()
@@ -596,6 +618,7 @@ def test_close_idempotent():
 
     manager.close()
     manager.close()
+    await_manager_shutdown(timeout=3)
 
     assert scheduler.shutdown.call_count == 1
 
@@ -640,6 +663,7 @@ def test_close_no_dispatcher_error():
     dispatcher.start()
 
     manager.close()
+    await_manager_shutdown(timeout=3)
 
     error_callback.assert_not_called()
 
@@ -651,6 +675,7 @@ def test_close_callbacks():
 
     manager.add_close_callback(callback)
     manager.close(reason="meep")
+    await_manager_shutdown(timeout=3)
 
     callback.assert_called_once_with(manager, "meep")
 
@@ -660,6 +685,7 @@ def test_close_blocking_scheduler_shutdown():
     scheduler = manager._scheduler
 
     manager.close()
+    await_manager_shutdown(timeout=3)
 
     scheduler.shutdown.assert_called_once_with(await_msg_callbacks=True)
 
@@ -669,6 +695,7 @@ def test_close_nonblocking_scheduler_shutdown():
     scheduler = manager._scheduler
 
     manager.close()
+    await_manager_shutdown(timeout=3)
 
     scheduler.shutdown.assert_called_once_with(await_msg_callbacks=False)
 
@@ -690,6 +717,7 @@ def test_close_nacks_internally_queued_messages():
     manager._messages_on_hold._messages_on_hold.append(messages[2])
 
     manager.close()
+    await_manager_shutdown(timeout=3)
 
     assert sorted(nacked_messages) == [b"msg1", b"msg2", b"msg3"]
 
--- a/tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
+++ b/tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
@@ -550,23 +550,16 @@ def make_running_manager(**kwargs):
     )
 
 
-def await_manager_shutdown(timeout=None):  # pragma: NO COVER
+def await_manager_shutdown(manager, timeout=None):
     # NOTE: This method should be called after manager.close(), i.e. after the shutdown
-    # thread has been started.
-    shutdown_thread = next(
-        (
-            thread
-            for thread in threading.enumerate()
-            if thread.name == streaming_pull_manager._REGULAR_SHUTDOWN_THREAD_NAME
-        ),
-        None,
-    )
+    # thread has been created and started.
+    shutdown_thread = manager._regular_shutdown_thread
 
-    if shutdown_thread is None:
-        return  # Shutdown already finished.
+    if shutdown_thread is None:  # pragma: NO COVER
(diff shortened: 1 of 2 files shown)


tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
```diff
@@ -550,6 +550,26 @@ def make_running_manager(**kwargs):
     )
 
 
+def await_manager_shutdown(timeout=None):  # pragma: NO COVER
+    # NOTE: This method should be called after manager.close(), i.e. after the shutdown
+    # thread has been started.
+    shutdown_thread = next(
```
pradn, line 556: I would rather we store the shutdown thread in a private variable in streaming_pull_manager and read it that way. It seems a bit cleaner than enumerating through all the threads, and is probably more performant, too.

tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
```diff
@@ -561,6 +574,7 @@ def test_close():
     ) = make_running_manager()
 
     manager.close()
+    await_manager_shutdown(manager, timeout=3)
```
jimfulton, line 577: Why not make this an argument to close?  `manager.close(wait=3)`

Then you wouldn't need await_manager_shutdown, nor the instance variable.

jimfulton (commented):
I'm not sure what flakiness this is fixing.

On my machine, the system test `test_publish_large_messages` still fails with a timeout on this branch. Maybe that's not what this PR is intended to address.

plamut:
@jimfulton It fixes the flakiness in _unit_ tests.

If a system test covering the same code is now flaky, too, we should investigate that as well, but it's a separate issue (that test depends on some assumption about how the backend works).

tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
plamut: There is another PubSub client called PubSub Lite, and it depends on some of the code from here - we try to be conservative with any interface changes, especially because we've just had a [regression](https://github.com/googleapis/python-pubsub/issues/318) when we added a new optional parameter.

We can discuss at the weekly if this particular change is safe to make.

jimfulton:
> @jimfulton It fixes the flakiness in _unit_ tests.
> 
> If a system test covering the same code is now flaky, too, we should investigate that as well, but it's a separate issue (that test depends on some assumption about how the backend works).

Oops, OK. :)

tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
jimfulton: My suggestion seems safe, as the omitted argument would behave as existing, but I'm happy to defer.

plamut: It indeed seems safe, but then again, we also though that the last time. :laughing:
I'll double check tomorrow, added to the agenda.

plamut: Oh, I should add - this class is internal, it's not exposed to the end users. Users can only trigger shutdown through the [StreamingPullFuture.cancel()](https://github.com/googleapis/python-pubsub/blob/c707f81fdadd118bd2bc48aebe74d453f35cad0d/google/cloud/pubsub_v1/subscriber/futures.py#L46-L51) method.

Now, we cannot touch that, because if the future sends something to the `manager.close()`, it will break PubSub Lite that uses the same future, but a different manager under the hood (i.e. one that does not support that new argument to `close()`).

jimfulton: Wouldn't that also be a problem if the future relied on the new method?

plamut: You mean if `StreamingPullFuture.cancel()` called something else instead, e.g. `self._manager.new_method()`? That would, too, be a problem, because the PubSub Lite's manager would not have that new method.

I might not understand the last question here? The future itself is not meant to be changed, at least not without coordination with PubSub Lite, meaning that changes to `StreamingPullManager.close()` would only serve the tests. Also, `close()` itself should be non-blocking, so that ` StreamingPullFuture.cancel()` calling it is also non-blocking.

Or is the suggestion to add a new for-test-only method to the manager, e.g. `manager.close_with_wait(wait=3)`?
(IMHO the test helper code should be palced in the tests themselves)

plamut: As discussed in the weekly - we'll add a "blocking close" method to the manager to be used as a helper in the tests, while `StreamingPullManager.close()` itself will remain intact (it needs to be non-blocking, and we cannot change `StreamingPullFuture.cancel()` that invokes it).

@jimfulton Did I summarize that correctly?

jimfulton: I guess. :)  I'm worried that the blocking close will duplicate code.  I'm fine with leaving the PR as it is.

jimfulton (approved):

PR #374, opened 2021-04-13
feat: make publish futures compatible with `concurrent.futures.as_completed()`

### LINKED ISSUE #368: Make publish futures compatible with `concurrent.futures.as_completed()`
#### Environment details

  - OS type and version: Linux
  - Python version: `Python 3.7.7`
  - pip version: `pip 21.0.1`
  - `google-cloud-pubsub` version: 
```Name: google-cloud-pubsub
Version: 2.3.0
Summary: Google Cloud Pub/Sub API client library
Home-page: https://github.com/googleapis/python-pubsub
Author: Google LLC
Author-email: googleapis-packages@google.com
License: Apache 2.0
Location: /Users/mike/.pyenv/versions/3.8.6/envs/detection-processor/lib/python3.8/site-packages
Requires: grpc-google-iam-v1, proto-plus, google-api-core, libcst
Required-by: 
```

#### Steps to reproduce

  1. Create multiple publish futures with `publisher.publish()`
  2. Use `in concurrent.futures.as_completed()` as a generator to trigger on completion.

#### Code example

```python
future_list = []

for msg in messages:
    future_list.append(publisher.publish(topic_path, msg))

for _ in concurrent.futures.as_completed(
    fs=future_list,
):
    # Do something
    continue
```

#### Stack trace
```
'Future' object has no attribute '_condition'
```

I understand that the publish future is not REALLY a concurrent future object. This attribute error does not happen with all messages, it's a minority of messages. I'm not sure what causes some publish futures to work with `as_completed` and some do not.

plamut:
Closes #368.

After studying the [as_completed()](https://github.com/python/cpython/blob/15ad30d88fead718b7eeff8c54454b82753d520e/Lib/concurrent/futures/_base.py#L201-L261) code from the standard library, it actually seemed easier to modify our futures implementation, as opposed to writing our own `as_completed()` logic, which would still require changes to the Future, too (think: locking and notifying).

The drawback of course is relying on implementation details from `concurrent.futures`, but good tests should help us here. Let's discuss this drawback here.

**Current approach pros:**
- Fewer code changes, no need to write our own `as_completed()` logic.
- The library's API surface does not change, no incompatibility with other official Pub/Sub clients, no dilemma and coordination on whether to move the implementation to `api-core`.

**Current approach cons:**
- Compatibility depends on `concurrent.futures` implementation details, although tests should help here. In addition, the standard lib's implementation seems very stable, and changes infrequent ([blame](https://github.com/python/cpython/blame/57873af35aad98c6428b1718aaee4b16a82ea3f5/Lib/concurrent/futures/_base.py)) - the key implementation details are more than 10 **years** old.

**PR checklist:**

### DIFF
--- a/google/cloud/pubsub_v1/futures.py
+++ b/google/cloud/pubsub_v1/futures.py
@@ -14,6 +14,7 @@
 
 from __future__ import absolute_import
 
+import concurrent.futures
 import threading
 import uuid
 
@@ -38,6 +39,14 @@ class Future(google.api_core.future.Future):
             with that model. The ``wait()`` and ``set()`` methods will be
             used. If this argument is not provided, then a new
             :class:`threading.Event` will be created and used.
+
+        condition (Optional[Any]): A condition variable, with the same interface as
+            :class:`threading.Condition`. This is provided so that callers
+            with different concurrency models (e.g. ``threading`` or
+            ``multiprocessing``) can supply a condition that is compatible
+            with that model. The ``acquire()`` and ``release()`` methods will be
+            used. If this argument is not provided, then a new
+            :class:`threading.Condition` will be created and used.
     """
 
     # This could be a sentinel object or None, but the sentinel object's ID
@@ -45,7 +54,7 @@ class Future(google.api_core.future.Future):
     # actually being a result.
     _SENTINEL = uuid.uuid4()
 
-    def __init__(self, completed=None):
+    def __init__(self, completed=None, condition=None):
         self._result = self._SENTINEL
         self._exception = self._SENTINEL
         self._callbacks = []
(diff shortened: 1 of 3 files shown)


google/cloud/pubsub_v1/futures.py
```diff
@@ -38,51 +39,69 @@ class Future(google.api_core.future.Future):
...
 
-        This still returns True in failure cases; checking :meth:`result` or
+        This still returns ``True`` in failure cases; checking :meth:`result` or
         :meth:`exception` is the canonical way to assess success or failure.
         """
-        return self._exception != self._SENTINEL or self._result != self._SENTINEL
+        with self._condition:
+            # The internal state is set to FINISHED when either of _exception or _result
+            # is set to a non-sentinel value, thus just checking the state is fine.
+            return self._state == concurrent.futures._base.FINISHED
```
plamut, line 104: Note to self - we probably don't need locking here.

plamut:
@pradn PTAL, if you have time.

It's true that the solution uses `concurrent.futures` implementation details, but the required changes are relatively small and we'll have tests to protect us.

Alternatively, we could provide our own `as_completed` helper, but that would likely require a bit more code, including the changes to our `Future` itself, plus we would be adding new stuff to the library surface that other clients probably don't have.

After reading the standard library implementation, the first approach actually seemed more favourable (IMHO).

pradn:
I can take a look later this week or early next week. :)

google/cloud/pubsub_v1/futures.py
```diff
@@ -140,47 +158,51 @@ def add_done_callback(self, callback):
...
+            self._state = concurrent.futures._base.FINISHED
+            self._completed.set()
 
-        # Set the result and trigger the future.
-        self._result = result
-        self._trigger()
+            for waiter in self._waiters:
+                waiter.add_result(self)
+
+        for callback in self._callbacks:
```
pradn, line 186: Do we not call callbacks on exception (with an exception parameter?)

pradn:
I took a first pass - looks reasonable. We need to check with Kamal if depending on the internals of a library class like this is fine. It seems to be helpful enough to the user to warrant this, but I wonder if so few are going to use this it might not be worth it.

google/cloud/pubsub_v1/futures.py
plamut: We do, there's a similar loop at the end of the `set_exception()` method. Or did I misunderstand the question?

plamut:
Using `as_completed()` is pretty standard when dealing with multiple futures and waiting on their completion, it's not that niche a use case IMHO.

IIRC the feature request has been confirmed (and Node is adding something similar, too), but I agree we should discuss the approach in this PR - I'll add it to the agenda.

google/cloud/pubsub_v1/futures.py
pradn: Oops, was mistaken via the diff UI.

pradn (dismissed):

pradn:
Looks good to me. It seems like we have consensus with the rest of the team, so perhaps simply ask for confirmation before merging?

plamut:
> Looks good to me. It seems like we have consensus with the rest of the team, so perhaps simply ask for confirmation before merging?

That's the plan.

And we might also borrow @jimfulton to have a say on using those implementation details, as he's been around Python ... probably since the fall of the Berlin Wall :laughing:

(which is likely not even far from the truth)

plamut:
After an offline discussion, we think relying on implementation details is not desirable, and the approach highest on the wish list is to swap the current implementation in the library with futures from `concurrent.futures` package.

The public interface is already compatible, but we need to see if that would break Pub/Sub Lite, and also determine the reason why the original authors decided to use their own implementation.

plamut:
Closing in favor of #397. We prefer less code and don't want to rely on the implementation details of `concurrent.futures.Future`.

PR #393, opened 2021-04-21
fix: do not crash if distribution cannot be found when extracting semver

hannahrogers-google:
In the case where a user is not running from within a pip package, the google-cloud-pubsub distribution might not be available. At the moment this will cause a fatal DistributionNotFound error because in the publisher and subscriber client, we try to extract the distribution version using 'pkg_resources.get_distribution("google-cloud-pubsub").version'.

In the case where a distribution is not found, making the change to set the version to 0.0.

### DIFF
--- a/google/cloud/pubsub_v1/publisher/client.py
+++ b/google/cloud/pubsub_v1/publisher/client.py
@@ -39,6 +39,8 @@ from google.pubsub_v1.services.publisher import client as publisher_client
 try:
     __version__ = pkg_resources.get_distribution("google-cloud-pubsub").version
 except pkg_resources.DistributionNotFound:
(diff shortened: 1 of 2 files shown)


pradn (changes requested):
Please add some context for when this call can throw.

pradn:
Change looks good! Can you please update the PR summary with some context?

hannahrogers-google:
> Change looks good! Can you please update the PR summary with some context?

Yes, I updated the PR summary and added a doc for context in the PR.  Thanks, Prad!

pradn:
Looks good, approved!

pradn (approved):

PR #396, opened 2021-04-23
chore(revert): revert preventing normalization

dandhlee:
reverts previous commit for preventing normalization of versioning

Reverts #382

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,20 +17,6 @@ import os
 
 import setuptools
 
-# Disable version normalization performed by setuptools.setup()
-try:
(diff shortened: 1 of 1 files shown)


plamut:
The samples failure is flakiness (500 on the backend).

plamut (commented):
@dandhlee  Same question as with the similar BigQuery PR - why do we need a revert, regression?

The revert itself looks good.

plamut:
Got an answer [here](https://github.com/googleapis/python-bigquery/pull/625#issuecomment-826882577).

plamut (approved):

PR #397, opened 2021-04-26
feat: make publish futures compatible with concurrent.futures.as_completed()

### LINKED ISSUE #368: Make publish futures compatible with `concurrent.futures.as_completed()`

plamut:
Closes #368.
Possibly supersedes #374.

This is a different approach to the same feature request which does not rely on implementation details of `concurrent.futures.Future` (as discussed offline). Instead, the custom base `Future` implementation is _replaced_ with the one from `concurrent.futures` package with only minor adjustments to fit specific needs of publisher/subscriber futures.

The interface stays the same, and the `isinstance()` checks should keep working, as `concurrent.futures.Future` is injected is injected into the right place in MRO.

**PR checklist**

### DIFF
--- a/google/cloud/pubsub_v1/futures.py
+++ b/google/cloud/pubsub_v1/futures.py
@@ -55,3 +55,19 @@ class Future(concurrent.futures.Future, google.api_core.future.Future):
         raise NotImplementedError(
             "Only used by executors from `concurrent.futures` package."
         )
+
+    def set_result(self, result):
+        """Set the return value of work associated with the future.
+
+        Do not use this method, it should only be used internally by the library and its
+        unit tests.
+        """
+        return super().set_result(result=result)
+
+    def set_exception(self, exception):
+        """Set the result of the future as being the given exception.
+
+        Do not use this method, it should only be used internally by the library and its
+        unit tests.
+        """
+        return super().set_exception(exception=exception)
--- a/google/cloud/pubsub_v1/futures.py
+++ b/google/cloud/pubsub_v1/futures.py
@@ -14,14 +14,13 @@
 
 from __future__ import absolute_import
 
-import threading
-import uuid
+import concurrent.futures
+import warnings
 
 import google.api_core.future
-from google.cloud.pubsub_v1.publisher import exceptions
 
 
-class Future(google.api_core.future.Future):
+class Future(concurrent.futures.Future, google.api_core.future.Future):
     """Encapsulation of the asynchronous execution of an action.
 
     This object is returned from asychronous Pub/Sub calls, and is the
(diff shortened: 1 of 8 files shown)


plamut:
@dpcollins-google Could this break Pub/Sub Lite in any way? Does the Lite client currently use any `Future` internals?

plamut:
```module 'concurrent.futures' has no attribute 'InvalidStateError'``` in Python 3.7 and below. Will refactor the affected unit tests.

File: google/cloud/pubsub_v1/futures.py [hunk 1]
```diff
@@ -31,156 +30,44 @@ class Future(google.api_core.future.Future):
...
     Args:
-        completed (Optional[Any]): An event, with the same interface as
-            :class:`threading.Event`. This is provided so that callers
-            with different concurrency models (e.g. ``threading`` or
-            ``multiprocessing``) can supply an event that is compatible
-            with that model. The ``wait()`` and ``set()`` methods will be
-            used. If this argument is not provided, then a new
-            :class:`threading.Event` will be created and used.
+        completed (Optional[Any]):
+             [deprecated] Not used anymore and will be removed in the future.
```
plamut, line 34: This should be fine, as `Future` objects should not be created by the users directly, and because we already did not really support concurrency model other than threading.

We only have the thread-based implementation of the `Batch` class, and the rest of the code (e.g. `Client.publish()` and `OrderedSequencer.publish()` ) already created `Future` instances using a default `completed`, which was `threading.Event()`.

plamut:
Status checks got stuck.

google/cloud/pubsub_v1/futures.py
jimfulton: Do we know of any clients that are passing this argument? (I see this package was one. :) )

Do we plan to get rid of this in the near future?

google/cloud/pubsub_v1/futures.py
```diff
@@ -31,156 +30,44 @@ class Future(google.api_core.future.Future):
...
+        unit tests.
         """
-        # Sanity check: A future can only complete once.
-        if self.done():
-            raise RuntimeError("set_result can only be called once.")
-
-        # Set the result and trigger the future.
-        self._result = result
-        self._trigger()
+        return super().set_result(result=result)
```
jimfulton, line 65: Why do we need this override?

google/cloud/pubsub_v1/futures.py
```diff
@@ -31,156 +30,44 @@ class Future(google.api_core.future.Future):
...
 
-        Args:
-            message_id (str): The message ID, as a string.
+        Do not use this method, it should only be used internally by the library and its
+        unit tests.
         """
-        self._completed.set()
-        for callback in self._callbacks:
-            callback(self)
+        return super().set_exception(exception=exception)
```
jimfulton, line 73: Why do we need this override?

google/cloud/pubsub_v1/publisher/futures.py
```diff
@@ -43,10 +57,4 @@ def result(self, timeout=None):
...
                 call execution.
         """
-        # Attempt to get the exception if there is one.
-        # If there is not one, then we know everything worked, and we can
-        # return an appropriate value.
-        err = self.exception(timeout=timeout)
-        if err is None:
-            return self._result
-        raise err
+        return super().result(timeout=timeout)
```
jimfulton, line 60: Why do we need this override?

jimfulton (approved):
This looks great.  I love the code reduction!

I'm a little nervous about the subclassing. 

I would minimize the number of overridden methods.

In `StreamingPullFuture`, I'd change the private attributes to use a double- rather than a single-underscore prefixes to get name mangling and greatly reduce the chance of a future conflict with the base class.

plamut:
> In `StreamingPullFuture`, I'd change the private attributes to use a double- rather than a single-underscore prefixes to get name mangling and greatly reduce the chance of a future conflict with the base class.

This is a very good suggestion, a case where we really do want to assure there's no accidental overriding.

> Why do we need this override? (x3)

To override the base implementations' docstrings', we want to prereserve the existing docstrings of Pub/Sub futures' methods. Do you know if there's a more idiomatic way to do this other than merely delegating the call to the implementation up the MRO chain?


> Do we know of any clients that are passing this argument? (I see this package was one. :) )
> 
> Do we plan to get rid of this in the near future?

The library only implements a thread-based batch, which is an [internal default](https://github.com/googleapis/python-pubsub/blob/3ee887cd0f8431e7c87db806a572491dd6216f97/google/cloud/pubsub_v1/publisher/client.py#L133) the publisher client uses. That batch class is the only thing that passed in the now-removed `completed` argument, and we are not aware of any 3rd parties that would considerably mangle the client code to change the publisher concurrency paradigm.

End users also do not instantiate the futures directly (and if they do, they violate [the contract](https://github.com/googleapis/python-pubsub/blob/3ee887cd0f8431e7c87db806a572491dd6216f97/google/cloud/pubsub_v1/futures.py#L30-L31) :) ).

No plans for removing the `completed` argument yet, but could make sense to do so, since since `concurrent.futures.Future` tracks its own state differently and doesn't need it.

jimfulton:
> > Why do we need this override? (x3)
> 
> To override the base implementations' docstrings', we want to prereserve the existing docstrings of Pub/Sub futures' methods. Do you know if there's a more idiomatic way to do this other than merely delegating the call to the implementation up the MRO chain?

IMO. It's not worth it.

The less overriding the better.


> 
> > Do we know of any clients that are passing this argument? (I see this package was one. :) )
> > Do we plan to get rid of this in the near future?
> 
> The library only implements a thread-based batch, which is an [internal default](https://github.com/googleapis/python-pubsub/blob/3ee887cd0f8431e7c87db806a572491dd6216f97/google/cloud/pubsub_v1/publisher/client.py#L133) the publisher client uses. That batch class is the only thing that passed in the now-removed `completed` argument, and we are not aware of any 3rd parties that would considerably mangle the client code to change the publisher concurrency paradigm.
> 
> End users also do not instantiate the futures directly (and if they do, they violate [the contract](https://github.com/googleapis/python-pubsub/blob/3ee887cd0f8431e7c87db806a572491dd6216f97/google/cloud/pubsub_v1/futures.py#L30-L31) :) ).
> 
> No plans for removing the `completed` argument yet, but could make sense to do so, since since `concurrent.futures.Future` tracks its own state differently and doesn't need it.

I would plan to remove it "soon". 

If we're confident that no clients are calling this, then I'd remove it now.

Every line of code we own is a liability -- overridden inherited code from external (to us) dependencies doubly (or triply) so.

These are suggestions. My approval stands.

google/cloud/pubsub_v1/subscriber/futures.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, manager):
         super(StreamingPullFuture, self).__init__()
         self._manager = manager
```
jimfulton, line 31: What about manager?

plamut: Uhm, what about it? :)
We need a reference to it to shut the stream down if a future gets cancelled.

If the question is about the concurrency model,  StreamingPullManager accepts a `scheduler` parameter (and uses a thread-based scheduler to dispatch messages to callbacks), but the class itself relies on threading synchronization primitives internally (non-configurable).

jimfulton: Why not `__manager`?

plamut: Oh, that... tunnel vision I guess, since that's an old already existing attribute. Will rename it.

jimfulton:
I have nothing more to say. :)

plamut:
I'll re-think about actually removing the `completed` argument right now when I re-visit this on Monday, since a loud error is probably better than silently changing the argument's effect to a no-op. Ditto for the other remark about overriding. Thanks for all the feedback!

dpcollins-google:
lgtm, shouldn't break pub/sub lite

anguillanneuf:
@plamut After this update is released, shall we update our publish/subscribe samples? Or add new samples to demonstrate how to work with these futures?

plamut:
On a second thought, removing the `completed` parameter is actually be better - it will cause a loud error for any code that uses it, instead of possibly introducing non-obvious bugs if its behavior was silently changed to a no-op. As for the overriding methods that preserve the existing docstrings and keep it under our control, I opted to keep them.

> After this update is released, shall we update our publish/subscribe samples? Or add new samples to demonstrate how to work with these futures?

@anguillanneuf  The existing samples should not need any modifications, since that's actually one of the goals - futures should continue working as before.

On the other hand, it would make sense to demonstrate their new capability, i.e. compatibility with `concurrent.futures.as_completed()`. Maybe as a use case where somebody publishes multiple messages in a loop and then wants to block until all operations complete.

Maybe something like the following?

```py
topic_path = publisher.topic_path(PROJECT_NAME, TOPIC_NAME)

future_to_msg = {}

for i in range(10):
    msg = f"message {i}".encode()
    suffix = "" if i % 2 else "x"
    future = publisher.publish(topic_path + suffix, msg)
    future_to_msg[future] = msg
    time.sleep(0.01)

for future in concurrent.futures.as_completed(future_to_msg):
    msg = future_to_msg[future].decode()
    try:
        result = future.result()
    except Exception as exc:
        print(f"Error publishing message {msg}: {exc}")
    else:
        print(f"Message successfully published: {msg}")
```

Sample output (with logging level set to CRITICAL to cut off background thread noise):
```
Message successfully published: message 5
Error publishing message message 0: 404 Resource not found (resource=topic-*****).
Error publishing message message 8: 404 Resource not found (resource=topic-*****).
Message successfully published: message 7
Message successfully published: message 9
Message successfully published: message 3
Message successfully published: message 1
Error publishing message message 6: 404 Resource not found (resource=topic-*****).
Error publishing message message 2: 404 Resource not found (resource=topic-*****).
Error publishing message message 4: 404 Resource not found (resource=topic-*****).
```

PR #406, opened 2021-04-30
fix: memory leak when publishing messages

### LINKED ISSUE #395: When I publish a data to pubsub topic, I have a memory leak on kubernetes pod.
Each time a public request is made, the memory in the pod of Kubernetes increases. Removing the publishing part does not increase memory. 
The pod is continuously restarted due to OOM.

#### Environment details

  - OS type and version: Debian GNU/Linux 10 
  - Python version: 3.8.7
  - pip version: 21.0.1
  - `google-cloud-pubsub` version: 2.4.1

#### Code example

```python

class PublishToPubsubCallback(Callback):
    def __init__(self, pod_name, project_id, topic_id, preprocess=[]):
        super(PublishToPubsubCallback, self).__init__(pod_name, preprocess=preprocess)
        self.project_id = project_id
        self.topic_id = topic_id
        self.publisher = pubsub_v1.PublisherClient()
        self.topic_path = self.publisher.topic_path(project_id, topic_id)
        
    def __call__(self, data):
        pub_data = json.dumps(data).encode('utf-8')
        self.publisher.publish(self.topic_path, pub_data)

callback = PublishToPubsubCallback(pod_name, project_id, topic_id)

async def run():
    while True:
        data = await watch_websocket_data()
        callback(data)

loop = asyncio.get_event_loop()
loop.run_until_complete(run())

```

스크린샷 2021-04-23 오후 4 58 16

plamut:
Fixes #395.

If publish threads are marked as daemonic, the leak seemingly disappears.

### How to verify

Set up a topic and a subscription, install `memoryprofiler`. Add the `@profile` decorator to the `main()` function from the [sample code](https://github.com/googleapis/python-pubsub/issues/395#issuecomment-829263348) and run it (with and without the fix):
```
$ mprof run sample.py
```

After each run, plot the memory usage graph:
```
$ mprof plot
```

The graphs should look similar to the ones posted [here](https://github.com/googleapis/python-pubsub/issues/395#issuecomment-829910303).

### PR checklist

### DIFF
--- a/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -207,9 +207,10 @@ class Batch(base.Batch):
 
     def _start_commit_thread(self):
         """Start a new thread to actually handle the commit."""
-
+        # NOTE: If the thread is *not* a daemon, a memory leak exists for some reason...
+        # https://github.com/googleapis/python-pubsub/issues/395#issuecomment-829910303
         commit_thread = threading.Thread(
-            name="Thread-CommitBatchPublisher", target=self._commit
+            name="Thread-CommitBatchPublisher", target=self._commit, daemon=True
         )
         commit_thread.start()
 
--- a/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -207,8 +207,9 @@ class Batch(base.Batch):
 
     def _start_commit_thread(self):
         """Start a new thread to actually handle the commit."""
-        # NOTE: If the thread is *not* a daemon, a memory leak exists for some reason...
+        # NOTE: If the thread is *not* a daemon, a memory leak exists sue to a CPython issue.
         # https://github.com/googleapis/python-pubsub/issues/395#issuecomment-829910303
+        # https://github.com/googleapis/python-pubsub/issues/395#issuecomment-830092418
         commit_thread = threading.Thread(
             name="Thread-CommitBatchPublisher", target=self._commit, daemon=True
         )
--- a/google/cloud/pubsub_v1/publisher/_batch/thread.py
+++ b/google/cloud/pubsub_v1/publisher/_batch/thread.py
@@ -207,7 +207,7 @@ class Batch(base.Batch):
 
     def _start_commit_thread(self):
         """Start a new thread to actually handle the commit."""
-        # NOTE: If the thread is *not* a daemon, a memory leak exists sue to a CPython issue.
+        # NOTE: If the thread is *not* a daemon, a memory leak exists due to a CPython issue.
         # https://github.com/googleapis/python-pubsub/issues/395#issuecomment-829910303
         # https://github.com/googleapis/python-pubsub/issues/395#issuecomment-830092418
         commit_thread = threading.Thread(
--- a/google/cloud/pubsub_v1/publisher/client.py
+++ b/google/cloud/pubsub_v1/publisher/client.py
@@ -375,8 +375,12 @@ class Client(object):
 
     def _start_commit_thread(self):
         """Start a new thread to actually wait and commit the sequencers."""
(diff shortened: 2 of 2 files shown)


plamut:
Server 500 when running samples, re-starting.

google/cloud/pubsub_v1/publisher/_batch/thread.py
```diff
@@ -207,9 +207,10 @@ def commit(self):
 
     def _start_commit_thread(self):
         """Start a new thread to actually handle the commit."""
-
+        # NOTE: If the thread is *not* a daemon, a memory leak exists for some reason...
+        # https://github.com/googleapis/python-pubsub/issues/395#issuecomment-829910303
```
plamut, line 211:
```suggestion
        # NOTE: If the thread is *not* a daemon, a memory leak exists sue to a CPython issue.
        # https://github.com/googleapis/python-pubsub/issues/395#issuecomment-829910303
        # https://github.com/googleapis/python-pubsub/issues/395#issuecomment-830092418
```

google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -375,8 +375,12 @@ def _ensure_commit_timer_runs_no_lock(self):
 
     def _start_commit_thread(self):
         """Start a new thread to actually wait and commit the sequencers."""
+        # NOTE: If the thread is *not* a daemon, a memory leak exists for some reason...
+        # https://github.com/googleapis/python-pubsub/issues/395#issuecomment-829910303
```
plamut, line 379:
```suggestion
        # NOTE: If the thread is *not* a daemon, a memory leak exists sue to a CPython issue.
        # https://github.com/googleapis/python-pubsub/issues/395#issuecomment-829910303
        # https://github.com/googleapis/python-pubsub/issues/395#issuecomment-830092418
```

plamut (commented):
Found a true cause of the issue, it's a bug in CPython.

google/cloud/pubsub_v1/publisher/_batch/thread.py
```diff
@@ -207,9 +207,11 @@ def commit(self):
 
     def _start_commit_thread(self):
         """Start a new thread to actually handle the commit."""
-
+        # NOTE: If the thread is *not* a daemon, a memory leak exists sue to a CPython issue.
```
plamut, line 210:
```suggestion
        # NOTE: If the thread is *not* a daemon, a memory leak exists due to a CPython issue.
```

google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -375,8 +375,13 @@ def _ensure_commit_timer_runs_no_lock(self):
 
     def _start_commit_thread(self):
         """Start a new thread to actually wait and commit the sequencers."""
+        # NOTE: If the thread is *not* a daemon, a memory leak exists sue to a CPython issue.
```
plamut, line 378:
```suggestion
        # NOTE: If the thread is *not* a daemon, a memory leak exists due to a CPython issue.
```

plamut (commented):
Fix typo

pradn:
The fix is good, but I wonder if we'd run into issues by setting these threads as [daemons](https://docs.python.org/3/library/threading.html#threading.Thread.daemon).

> The entire Python program exits when no alive non-daemon threads are left.

So, this would mean that scripts that may have created a publisher client, published a bunch of messages, and had only exited when the commit threads (non-daemon) finished running - these scripts could now terminate earlier. However, this may not be a problem, because if users want to guarantee publishes happened, they need to wait for the publish futures to finish with success.

Anyway, just wondering what your thoughts were here.

pradn:
Peter and I spoke offline: possible premature script termination b/c the threads are no daemons doesn't break a customer guarantee, because as long as the user hasn't received a completed future, they can't assume the publish succeeded.

pradn (approved):
LGTM

plamut:
Since it's already late into the workweek, I'll cut a release containing this on Monday.

PR #416, opened 2021-05-21
fix: ACK deadline set for received messages can be too low 

### LINKED ISSUE #413: The initial ACK deadline set for the messages received could be too low
Is there a reason that https://github.com/googleapis/python-pubsub/blob/master/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py#L493 calls `self.ack_histogram.percentile(99)` directly rather than using https://github.com/googleapis/python-pubsub/blob/master/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py#L225 which is what the Leaser thread uses for how long it sleeps between updates?

I think this is causing a bug where the the average message time decrease and so the 99th percentile decreases quicker than the "sticky" `ack_deadline` method. The initial ack_deadline is then shorter than the time between Leaser refreshes and the messages expire before the lease is renewed.

plamut:
Fixes #413.

This PR makes sure that the ACK deadline set for the received messages is always consistent with what the leaser uses internally when extending the ACK deadlines for the leased messages.

See the issue description and a [comment](https://github.com/googleapis/python-pubsub/issues/413#issuecomment-844995852) explaining a possible sequence of events that lead to a bug.

**PR checklist**

### DIFF
--- a/google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
+++ b/google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
@@ -128,7 +128,9 @@ class Leaser(object):
             # Determine the appropriate duration for the lease. This is
             # based off of how long previous messages have taken to ack, with
             # a sensible default and within the ranges allowed by Pub/Sub.
-            deadline = self._manager.ack_deadline
+            # Also update the deadline currently used if enough new ACK data has been
+            # gathered since the last deadline update.
+            deadline = self._manager._obtain_ack_deadline(maybe_update=True)
             _LOGGER.debug("The current deadline value is %d seconds.", deadline)
 
             # Make a copy of the leased messages. This is needed because it's
--- a/google/cloud/pubsub_v1/subscriber/_protocol/histogram.py
+++ b/google/cloud/pubsub_v1/subscriber/_protocol/histogram.py
@@ -15,6 +15,10 @@
 from __future__ import absolute_import, division
 
 
+MIN_ACK_DEADLINE = 10
+MAX_ACK_DEADLINE = 600
+
+
 class Histogram(object):
     """Representation of a single histogram.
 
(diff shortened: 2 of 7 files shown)


plamut:
Hmm... while this PR fixes the issue caused by directly using the `p99` value that can be different from `manager.ack_deadline` used by the leaser, an ACK deadline can still expire.

Consider - `manager.ack_deadline` is currently 600, because processing on of the messages took 1000 =  minutes. The leaser goes to sleep for `0.9 * 600 = 540` seconds. While the leaser is asleep, more messages are received and processed quickly, dropping the histogram's `p99` value to 10 seconds, consequently adjusting the value of `manager._ack_deadline`, too.

When yet more messages arrive, the shortened ACK deadline will be used for them **while the leaser is still asleep** (due to the previous "long" deadline). The leaser will not wake in time to extend the deadlines of these new messages.

Bottom line - the ACK deadline should probably not change while the leaser is asleep so that `_on_response()` does not use a deadline that could be shorter than the current sleep period of the leaser.

plamut:
Fixed, only the leaser now updates the `manager.ack_deadline` value. This means that the leased messages will never use an ACK deadline that is shorter than the duration of the leaser's current sleep period.

google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
```diff
@@ -128,7 +128,9 @@ def maintain_leases(self):
             # Determine the appropriate duration for the lease. This is
             # based off of how long previous messages have taken to ack, with
             # a sensible default and within the ranges allowed by Pub/Sub.
-            deadline = self._manager.ack_deadline
+            # Also update the deadline currently used if enough new ACK data has been
+            # gathered since the last deadline update.
+            deadline = self._manager._obtain_ack_deadline(maybe_update=True)
```
jimfulton, line 133: What's the rationale for updating the ack deadline here?

Why not in the dispatcher ack method after updating the histogram?

plamut: If an ACK deadline is updated while the leaser is sleeping (let's assume it gets shorter), newly arrived messages will use that updated deadline. This could be problematic if processing of these messages gets delayed - the (new) deadline would expire, because the leaser would still be sleeping, as it went to sleep using the older (longer) deadline.

It's actually the very issue this PR fixes.

jimfulton: IOW, we can't _decrease_ the deadline while the leaser is sleeping, because then it wouldn't wake in time to renew, as it would be sleeping based on the old longer deadline.

jimfulton (approved):

google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
plamut: Correct, you summarized it well.

PR #422, opened 2021-06-02
fix: threads can skip the line in publisher flow controller

### LINKED ISSUE #421: A thread might skip the line in publisher flow controller
In publisher `FlowController`, the checks whether or not a message can go through based on the current message count are not strict enough.

When a thread blocks because the message limit is hit, it [goes to sleep](https://github.com/googleapis/python-pubsub/blob/e907f6e05f59f64a3b08df3304e92ec960997be6/google/cloud/pubsub_v1/publisher/flow_controller.py#L145) in the `add()` method until a slot for another message becomes available. This happens when publishing one of the earlier messages completes and that message is released from the flow controller. Assuming that there's now again enough capacity to accept at least one new message, the sleeping threads get notified.

Now, just when our sleeping thread wakes up, but before it re-acquires the lock, another thread might arrive, call `add()` and acquire the lock just in front of the the woken up thread's nose. The new thread sees that there's room for a new message and takes it, forcing the older thread to wait until the next opportunity.

**Expected result:**
Threads should always be let through in FIFO order.

The fix is to force threads to reserve a slot for the message in the same way they need to reserve the free bytes capacity. Also, any released message slots must be distributed in FIFO order among the queued up threads.

In an extreme case, this issue can also cause some starving, though that becomes increasingly unlikely, as the waiting thread will eventually get lucky and grab the lock before the threads arriving after it will.

plamut:
Fixes #421.

Contrary to the bytes reservations, free message slots are not distributed in FIFO order among the waiting threads, meaning that a message arriving later could be accepted before a message already waiting in the queue.

This PR fixes it by enforcing FIFO distribution of available message slots + some simplifications of the logic and data structures.

**Tip:** Might be easier to review commit-by-commit, each contains a well-rounded change to the codebase.

**PR checklist:**

### DIFF
--- a/google/cloud/pubsub_v1/publisher/flow_controller.py
+++ b/google/cloud/pubsub_v1/publisher/flow_controller.py
@@ -193,65 +193,54 @@ class FlowController(object):
                 self._message_count = max(0, self._message_count)
                 self._total_bytes = max(0, self._total_bytes)
 
-            # TODO: merge these two together to only loop through threads once...
-            # And distribute both bytes and slots at the same time
-            self._distribute_available_slots()
-            self._distribute_available_bytes()
+            self._distribute_available_capacity()
 
             # If at least one thread waiting to add() can be unblocked, wake them up.
             if self._ready_to_unblock():
                 _LOGGER.debug("Notifying threads waiting to add messages to flow.")
                 self._has_capacity.notify_all()
 
-    def _distribute_available_slots(self):
-        """Distribute available message slots among the waiting threads in FIFO order.
+    def _distribute_available_capacity(self):
+        """Distribute available capacity among the waiting threads in FIFO order.
 
         The method assumes that the caller has obtained ``_operational_lock``.
         """
-        available = (
+        available_slots = (
             self._settings.message_limit - self._message_count - self._reserved_slots
         )
+        available_bytes = (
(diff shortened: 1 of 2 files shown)


pradn:
The solution looks good, but I worry about the increase in complexity. Do you think the risk of starvation is high enough to warrant this change?

plamut:
>  Do you think the risk of starvation is high enough to warrant this change?

I'd say yes, here's my thinking:
- Any reported starvation bugs could be hard to verify/reproduce consistently, thus better to get rid of this now.
- Ruby and Java do the same, they both control free slot distribution (in FIFO order).
- This complexity is self-contained within an independent component, making it more manageable. Besides, some of the logic has also been simplified, so the net complexity increase is not that big (IMHO). :slightly_smiling_face: 

An argument against could be that users will not even be publishing messages in multiple threads using a shared client instance, most of them will probably just publish them in a single thread in a tight loop. But since we decided back then to make the flow controller thread-safe, then let's also keep it robust, which requires some inherent complexity anyway.

pradn:
Okay, fwiw I agree with you that it's best to prevent hard-to-debug issues now. I wish there was a standard flow control mechanism we could just pull off the shelf, because there's some subtlety in this class.

pradn (approved):

PR #424, opened 2021-06-03
docs: block until the streaming pull shuts down

### LINKED ISSUE #423: Race conditions in streaming pull samples
The streaming pull future's `cancel()` is non-blocking. If subscriber client is used as a context manager that closes the underlying channel, but the client does not block when shutting down the stream (e.g. it just calls `future.cancel()`), the channel can be closed before the subscriber sends the last requests to the server (e.g. auto NACK of any remaining messages).

The block below is common in the existing samples:
```py
with subscriber:
    try:
        # When `timeout` is not set, result() will block indefinitely,
        # unless an exception occurs first.
        streaming_pull_future.result(timeout=timeout)
    except TimeoutError:
```

The samples should demonstrate how to block before leaving the `with` block, but they currently do no, and users have observed errors on shutdown ([example](https://github.com/googleapis/python-pubsub/issues/176#issuecomment-850357713)). This should be fixed.

plamut:
Fixes #423.

If subscriber client is used as a context manager, we need to block until the shutdown is complete before leaving the `with` block. See the issue description for more details.

**PR checklist:**

### DIFF
--- a/google/cloud/pubsub_v1/subscriber/client.py
+++ b/google/cloud/pubsub_v1/subscriber/client.py
@@ -188,7 +188,8 @@ class Client(object):
             try:
                 future.result()
             except KeyboardInterrupt:
-                future.cancel()
+                future.cancel()  # Trigger the shutdown.
+                future.result()  # Block until the shutdown is complete.
 
         Args:
             subscription (str): The name of the subscription. The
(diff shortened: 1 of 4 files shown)


pradn (approved):
Looks good, thanks!

PR #425, opened 2021-06-10
test: presubmit against Lite samples

anguillanneuf:
Add a presubmit test to run against samples in https://github.com/googleapis/python-pubsublite.

This is to prevent future development here from breaking `python-pubsublite` unknowingly, which has happened before in version 2.4.0: https://pypi.org/project/google-cloud-pubsub/#history

Hard-coded `pytest` code instead of using `nox`.

### DIFF
--- a/.kokoro/presubmit-against-pubsublite-samples.sh
+++ b/.kokoro/presubmit-against-pubsublite-samples.sh
@@ -1,0 +1,105 @@
+#!/bin/bash
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# `-e` enables the script to automatically fail when a command fails
+# `-o pipefail` sets the exit code to the rightmost comment to exit with a non-zero
+set -eo pipefail
+# Enables `**` to include files nested inside sub-folders
+shopt -s globstar
+
+# Disable buffering, so that the logs stream through.
+export PYTHONUNBUFFERED=1
+
+# Debug: show build environment
+env | grep KOKORO
+
+# Install nox
+python3.6 -m pip install --upgrade --quiet nox
+
+# Use secrets acessor service account to get secrets
+if [[ -f "${KOKORO_GFILE_DIR}/secrets_viewer_service_account.json" ]]; then
+    gcloud auth activate-service-account \
+	   --key-file="${KOKORO_GFILE_DIR}/secrets_viewer_service_account.json" \
+	   --project="cloud-devrel-kokoro-resources"
+fi
+
+cd github/python-pubsub
+
+# This script will create 3 files:
+# - testing/test-env.sh
+# - testing/service-account.json
+# - testing/client-secrets.json
+./scripts/decrypt-secrets.sh
+
+source ./testing/test-env.sh
+export GOOGLE_APPLICATION_CREDENTIALS=$(pwd)/testing/service-account.json
+
+# For cloud-run session, we activate the service account for gcloud sdk.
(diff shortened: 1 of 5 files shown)


anguillanneuf:
The test was run and passed. See [`fusion:prod:cloud-devrel/client-libraries/python/googleapis/python-pubsub/presubmit/presubmit-against-pubsublite-samples`](https://fusion.corp.google.com/projectanalysis/summary/KOKORO/prod:cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-pubsub%2Fpresubmit%2Fpresubmit-against-pubsublite-samples)

Highlighting a few areas for reviewers to comment:
- For some reason, the added presubmit test isn't showing up in the build details by default, even though it ran. Also, a passing `cloud-devrel/client-libraries/python/googleapis/python-pubsub/presubmit/presubmit` seems to overwrite a failing `cloud-devrel/client-libraries/python/googleapis/python-pubsub/presubmit/presubmit-against-pubsublite-samples`. What can I do there?
- I've hard-coded the test to use Python 3.6. Perhaps I shall add 3.7 and 3.8 too?

busunkim96:
> For some reason, the added presubmit test isn't showing up in the build details by default, even though it ran. Also, a passing cloud-devrel/client-libraries/python/googleapis/python-pubsub/presubmit/presubmit seems to overwrite a failing cloud-devrel/client-libraries/python/googleapis/python-pubsub/presubmit/presubmit-against-pubsublite-samples. What can I do there?

I think if you rename the check on the internal config side it will show up as a separate check and not overwrite the existing presubmit.

> I've hard-coded the test to use Python 3.6. Perhaps I shall add 3.7 and 3.8 too?

Do you think it's possible for a breaking change to only be noticeable in one Python version but not the others? If so, you can add the others. If you think it's unlikely I think it's OK to just have 3.6. :)

.kokoro/presubmit-against-pubsublite-samples.sh
```diff
@@ -0,0 +1,105 @@
...
+
+    # Use pytest to execute tests for py-3.6
+    python3.6 -m venv py-3.6
+    source py-3.6/bin/activate
+    # Install python-pubsublite samples tests requirements.
+    python -m pip install -r requirements.txt -q
+    python -m pip install -r requirements-test.txt -q
+    # Install python-pubsub from source.
+    python -m pip install -e "$ROOT" -q
+    python -m pytest quickstart_test.py
```
busunkim96, line 86: Just checking that you only want to run the quickstart

busunkim96 (approved):
Once the Kokoro status check has been renamed please add it to the repo settings config so it is marked required.  https://github.com/googleapis/python-pubsub/blob/master/.github/sync-repo-settings.yaml#L7

.kokoro/presubmit-against-pubsublite-samples.sh
anguillanneuf: Yes, it should provide good coverage.

anguillanneuf:
Thank you! Internal change (cl/378896781) submitted.

> Do you think it's possible for a breaking change to only be noticeable in one Python version but not the others?
It's unlikely AFAIK. Having just 3.6 should be enough.

.kokoro/presubmit-against-pubsublite-samples.sh
```diff
@@ -0,0 +1,105 @@
...
+    deactivate py-3.6
+    rm -rf py-3.6/
+
+    EXIT=$?
```
plamut, line 90: Maybe I'm missing something, but wouldn't this only check the status code of the last command, i.e. `rm -rf py-3.6/` ? Shouldn't this be placed right after the `python -m pytest quickstart_test.py` line instead?

plamut (commented):
This check is very welcome, thanks for adding it!

I see that it has already been merged, but please double check if we gather the status code at the right place?

.kokoro/presubmit-against-pubsublite-samples.sh
plamut: cc: @anguillanneuf @busunkim96

busunkim96: Ah that sounds right to me. @anguillanneuf Could you move up the `EXIT=$?` line to after `python -m pytest ...`?

anguillanneuf: Thank you both for catching that. https://github.com/googleapis/python-pubsub/pull/427

PR #476, opened 2021-08-24
samples: add type hints to all samples

### LINKED ISSUE #264: samples: add type hints to all samples

plamut:
Closes #264.

And here they are, annotated samples.

We should follow up with a `pytype` CI check, I'll open a separate issue for that.

**PR checklist:**

### DIFF
--- a/samples/snippets/iam.py
+++ b/samples/snippets/iam.py
@@ -24,7 +24,7 @@ at https://cloud.google.com/pubsub/docs.
 import argparse
 
 
-def get_topic_policy(project_id, topic_id):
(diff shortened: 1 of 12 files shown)


anguillanneuf (approved):
LGTM! Thank you Peter!

plamut:
I will follow up with adding `pytype` checks to leverage these type annotations and the ones we'll add to the code in the future (#477).

PR #495, opened 2021-09-03
process: set publisher methods' max retry to 600s

### LINKED ISSUE #494: Set default publisher retry and deadline settings to be consistent with desired values
Similar to the [java-pubsub PR](https://github.com/googleapis/java-pubsub/pull/820), the current default retry and timeout settings for publisher methods should be adjusted.

The methods are [wrapped](https://github.com/googleapis/python-pubsub/blob/8ae2b5a755b5240b7b9d1f7bd6135c8c833973e3/google/pubsub_v1/services/publisher/transports/base.py#L160-L303) in the base publisher transport. It's a generated part of the library, and the generator uses [these settings](https://github.com/googleapis/googleapis/blob/9c7eb1f9a78060a15cc309ab51ee0600c2e6f4f7/google/pubsub/v1/pubsub_grpc_service_config.json#L2-L100).

It appears that the settings are consistent with the values shared in internal docs, except for the "total timeout", which is the retry deadline, i.e. _"how long to keep retrying in seconds"_.

plamut:
Fixes #494.

**PR checklist:**


tseaver (commented):
This change, while expedient, is a bit ugly:  the deadlines are emitted from [this template](https://github.com/googleapis/gapic-generator-python/blob/630bff179a7a5babccd4bf747ed7cd9950bc3b8e/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/transports/base.py.j2#L177), based on this [method configuration](https://github.com/googleapis/googleapis/blob/9c7eb1f9a78060a15cc309ab51ee0600c2e6f4f7/google/pubsub/v1/pubsub_grpc_service_config.json#L88) in the `googleapis` repository.

If there is a general usecase for setting the deadlines to a higher multiple of the proto-repository-specified timeout, then we should have an issue open against `gapic-generator-python` to support it (and maybe we need additional configuration available in the service JSON config).

plamut:
Yeah, the thing is that currently the same value is used for retry deadline and the default timeout, thus indeed a generator change would be required to support it.

I can check at the next Python libs weekly if this is something other libraries could use as well, or whether it's just something specific to Pub/Sub clients - @kamalaboulhosn do you maybe know from the top of your head?

plamut:
**Update:** As discussed offline, we will go with this patchy version until it is decided where (and if) in the generator should this be changed.

kamalaboulhosn:
Yeah, this does look a little hacky. I guess this is something that comes from the service config. @hongalex is this a config we can change in a single place that can be distributed to the repos so we don't to do these ad-hoc changes?

plamut:
The problem here is that the gapic code generator currently [uses the same setting](https://github.com/googleapis/gapic-generator-python/blob/630bff179a7a5babccd4bf747ed7cd9950bc3b8e/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/transports/base.py.j2#L177-L182) for both the timeout and the retry deadline.

FWIW, it was agreed on the Python libs weekly yesterday that it needs to be investigated whether this is a quirk in the generator, or a lack of feature that would have to be implemented. Until then, the OwlBot replacement rule should be fine in order to not block this P1 change for too long.

kamalaboulhosn (approved):

PR #502, opened 2021-09-20
chore: relocate owl bot post processor

SurferJeffAtGoogle:
chore: relocate owl bot post processor


plamut (commented):
The OwlBot check fails with status CANCELLED. I'll try to re-run it, but if it occurs again somebody with access to internals logs will have to investigate.

(and that samples failure is just flakiness)

PR #542, opened 2021-11-23
process: make sure type annoations pass with mypy

### LINKED ISSUE #536: Make sure type checks pass with mypy as well
#500 added type annotation to the codebase, but we only checked that static type analysis passes under `pytype`. A lot of people, on the other hand, use `mypy`, but `mypy` currently reports 97 errors.

If we want users to leverage our type annotations, we should make sure that the checks pass with `mypy` as well, and then we can also add `py.typed` file to declare the library type-checked.

An additional requirement of this issue is that a new nox session named `mypy` needs to be added and enabled by default.

plamut:

This is a draft PR. The new noxfile check, mypy, passes locally, but there are still a few things to consider:
 - [x] The generated code is omitted from type checks, because `mypy` reports several errors for it
   (will be done separately, since it's currently blocked by https://github.com/googleapis/gapic-generator-python/issues/1092 )
 - [x] ~~The PR adds `py.typed` marker file, but we might want to postpone this addition until the generated code is error-free.~~   Marker file removed for the time being.
 - [x] Some type aliases such as `_TimeoutType` are defined in multiple places. We should probably extract the definition to a single central place within the library.

**PR checklist:**

### DIFF
--- a/google/cloud/__init__.py
+++ b/google/cloud/__init__.py
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import List
+
 try:
     import pkg_resources
 
@@ -21,4 +23,6 @@ try:
 except ImportError:
     import pkgutil
 
-    __path__ = pkgutil.extend_path(__path__, __name__)
+    __path__: List[str] = pkgutil.extend_path(
+        __path__, __name__
+    )  # pytype: disable=annotation-type-mismatch
--- a/google/cloud/__init__.py
+++ b/google/cloud/__init__.py
@@ -23,6 +23,4 @@ try:
 except ImportError:
     import pkgutil
 
-    __path__: List[str] = pkgutil.extend_path(
-        __path__, __name__
-    )  # pytype: disable=annotation-type-mismatch
+    __path__: List[str] = pkgutil.extend_path(__path__, __name__)  # type: ignore
--- a/google/cloud/pubsub_v1/publisher/_sequencer/ordered_sequencer.py
+++ b/google/cloud/pubsub_v1/publisher/_sequencer/ordered_sequencer.py
@@ -290,12 +290,12 @@ class OrderedSequencer(sequencer_base.Sequencer):
         """
         with self._state_lock:
             if self._state == _OrderedSequencerStatus.PAUSED:
-                future: Optional[futures.Future] = futures.Future()
+                errored_future = futures.Future()
                 exception = exceptions.PublishToPausedOrderingKeyException(
                     self._ordering_key
                 )
-                cast(futures.Future, future).set_exception(exception)
-                return cast(futures.Future, future)
+                errored_future.set_exception(exception)
+                return errored_future
 
             # If waiting to be cleaned-up, convert to accepting messages to
             # prevent this sequencer from being cleaned-up only to have another
--- a/google/cloud/pubsub_v1/publisher/_sequencer/ordered_sequencer.py
+++ b/google/cloud/pubsub_v1/publisher/_sequencer/ordered_sequencer.py
@@ -16,7 +16,7 @@ import enum
 import collections
 import threading
 import typing
-from typing import cast, Deque, Iterable, Optional, Sequence, Union
+from typing import cast, Deque, Iterable, Optional, Sequence
 
 from google.api_core import gapic_v1
 from google.cloud.pubsub_v1.publisher import futures
@@ -26,12 +26,10 @@ from google.cloud.pubsub_v1.publisher._batch import base as batch_base
 from google.pubsub_v1 import types as gapic_types
 
 if typing.TYPE_CHECKING:  # pragma: NO COVER
+    from google.cloud.pubsub_v1 import types
     from google.cloud.pubsub_v1.publisher import _batch
     from google.cloud.pubsub_v1.publisher.client import Client as PublisherClient
     from google.pubsub_v1.services.publisher.client import OptionalRetry
-
-
-_TimeoutType = Union[gapic_types.TimeoutType, gapic_v1.method._MethodDefault]
 
 
 class _OrderedSequencerStatus(str, enum.Enum):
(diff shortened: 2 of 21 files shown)


google/cloud/__init__.py
```diff
@@ -14,11 +14,15 @@
...
     import pkg_resources
 
     pkg_resources.declare_namespace(__name__)
 except ImportError:
     import pkgutil
 
-    __path__ = pkgutil.extend_path(__path__, __name__)
+    __path__: List[str] = pkgutil.extend_path(
+        __path__, __name__
+    )  # pytype: disable=annotation-type-mismatch
```
tseaver, line 28: Given that `mytype` and `pytype` disagree, I think I'd prefer just `# type: ignore` here.  *Nobody* is going to care in the wild, anyway.

google/cloud/pubsub_v1/publisher/_sequencer/ordered_sequencer.py
```diff
@@ -289,12 +290,12 @@ def publish(
         """
         with self._state_lock:
             if self._state == _OrderedSequencerStatus.PAUSED:
-                future = futures.Future()
+                future: Optional[futures.Future] = futures.Future()
```
tseaver, line 293: We could avoid the `Optional` and `cast` here by using a different variable name.

google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -380,7 +384,10 @@ def on_publish_done(future):
                 if retry is gapic_v1.method.DEFAULT:
                     # use the default retry for the publish GRPC method as a base
                     transport = self.api._transport
-                    retry = transport._wrapped_methods[transport.publish]._retry
+                    retry = typing.cast(
```
tseaver, line 387: We might avoid the `cast` here, too, by using a different variable name, e.g., `base_retry`.

google/cloud/pubsub_v1/publisher/flow_controller.py
```diff
@@ -25,6 +25,9 @@
 _LOGGER = logging.getLogger(__name__)
 
 
+MessageType = Type[types.PubsubMessage]  # type: ignore  # pytype: disable=module-attr
```
tseaver, line 28: I don't understand the `type: ignore` here, and I think this may be a misuse of `Type[]`.

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -120,17 +121,38 @@ def dispatch_callback(self, items: Sequence[RequestItem]) -> None:
         _LOGGER.debug("Handling %d batched requests", len(items))
 
         if batched_commands[requests.LeaseRequest]:
-            self.lease(batched_commands.pop(requests.LeaseRequest))
+            lease_requests = cast(
+                List[requests.LeaseRequest],
+                batched_commands.pop(requests.LeaseRequest),
+            )
+            self.lease(lease_requests)
```
tseaver, line 128: The typing contortions here are a sign that using a single `batched_commands` mapping is a broken design.  It would be far cleaner to keep four separate lists.

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -180,7 +203,8 @@ def lease(self, items: Sequence[requests.LeaseRequest]) -> None:
         Args:
             items: The items to lease.
         """
-        self._manager.leaser.add(items)
+        leaser = cast("Leaser", self._manager.leaser)
```
tseaver, line 206: Seems like a signal that `StreamingPullManager.leaser` needs better typing.  If it is `Optional`, then an `assert self._manager.leaser is not None` would work here.

google/cloud/pubsub_v1/subscriber/_protocol/leaser.py
```diff
@@ -183,7 +194,8 @@ def maintain_leases(self) -> None:
                 #       without any sort of race condition would require a
                 #       way for ``send_request`` to fail when the consumer
                 #       is inactive.
-                self._manager.dispatcher.modify_ack_deadline(
+                dispatcher = cast("Dispatcher", self._manager.dispatcher)
```
tseaver, line 197: Again, the `cast` looks suspicious to me.

tseaver (commented):
Overall, I'd like to see us using only one typing checker, and I think 'mytype' is the consensus favorite.

google/cloud/pubsub_v1/publisher/flow_controller.py
plamut: The ignore here is needed, because `PubsubMessage` is dynamically injected into `google/cloud/pubsub_v1/types.py` and both type checkers think it does not exist. The alias was created to not repeat the same ignore comment in every line where `message` is a method parameter.

Using a plain `types.PubsubMessage` in the alias results in `mypy` complaining that variable `MessageType` is not valid as a type. Using `Type[...]` works around that.

(alternatives welcome, especially if they are canonical)

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
plamut: Leaser can be `None`, thus the `cast()` to avoid the warning. But I like the `assert` trick better, it's much more clear what's going on here.

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
plamut: Fair point, will refactor.

plamut:
A few trivial errors, will fix the checks tomorrow.

tseaver (approved):

plamut:
@tseaver The required changes turned out to be more than just a few trivial lines, thus please take another look when you manage.

It's mostly just getting rid of `cast()` for Optional types, and one extra case in Dispatcher had to be covered after splitting `batch_commands` into separate lists (coverage complaint).

tseaver (approved):

PR #546, opened 2021-12-02
process: persist mypy session in noxfile

### LINKED ISSUE #545: Persist mypy nox session
The `mypy` session was manually added to the noxfile, which is auto-generated, thus OwlBot overrides those changes (https://github.com/googleapis/python-pubsub/pull/544).

Instead, the changes should be injected into `noxfile.py` through `owlbot.py` replacements, at least until `mypy` session is not added by the code generator itself.

plamut:
Fixes #545.

**PR checklist:**


parthea (approved):

plamut:
Thanks for the quick review, @parthea !

PR #550, opened 2021-12-13
feat: exactly-once delivery support

pradn:
* Add new field to SubscriptionProperties and modify the GAPIC layer with just this one change. This change will come from the auto-GAPIC generated code once the proto change is public.
* Add async ack/modack/nack methods to the Message class. Exactly-once users have to use this to get the status of ack/modack/nack requests. The futures returned by these methods may be completed by the AcknowledgeError exception (with detailed AcknowledgeStatus status code) or a success AcknowledgeStatus code. These futures are of a new type, one used only on the subscriber-side for propagating exactly-once errors. This future mirrors the publishing-side future that already exists. The futures are completed immediately if exactly-once delivery is not enabled, upon successful ack/modack completion, and when errors/exceptions occur.
* Split unary ack/modack methods into two in the subscriber manager. Some repeated code but it's simpler this way.
* Keep track of whether exactly-once is enabled for a subscription by looking at the subscription's SubscriptionProperties, returned in the streaming response.
* Set the minimum ack deadline to 60 secs if exactly-once is known to be turned on. This deadline is used by the leaser.
* Add new min-lease-extension parameter. If the user sets this, it overrides the auto-set param based on whether exactly-once is enabled or not.
* Retry unary ack/modacks if we get errors in the GRPC call's ErrorInfo (no retry delays as of yet since the server will eventually fail them if the ack expires). This is done in the dispatcher, which retries only the acks/modacks that have failed temporarily. Permanent failures are surfaced to the user as future exceptions. These retries use an exponential backoff with a max of 10 min, which is sufficient since the server allows ack/modacks for 10 mins max. These retries are triggered on a new thread to prevent the dispatcher thread from being blocked by retry-wait sleeps.
* Retry lease modacks, both immediately upon receipt of messages and also in the background. If retries fail or if permanent errors arise, we log the errors but don't otherwise disrupt user code.
* Send a new ack deadline via stream_ack_deadline_seconds whenever the exactly_once setting for a stream changes.
* Add tests ensuring 100% coverage
* Usage sample for this feature added at https://github.com/googleapis/python-pubsub/pull/588

### DIFF
--- a/google/cloud/pubsub_v1/proto/pubsub.proto
+++ b/google/cloud/pubsub_v1/proto/pubsub.proto
@@ -1164,6 +1164,7 @@ message StreamingPullRequest {
 message StreamingPullResponse {
   // Subscription properties sent as part of the response.
   message SubscriptionProperties {
+    bool exactly_once_delivery_enabled = 1;
     // True iff message ordering is enabled for this subscription.
     bool message_ordering_enabled = 2;
   }
--- a/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
+++ b/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
@@ -70,11 +70,11 @@ IDs at a time.
 
 _MIN_EXACTLY_ONCE_DELIVERY_ACK_MODACK_RETRY_DURATION_SECS = 1
 """The time to wait for the first retry of failed acks and modacks when exactly-once
-is enabled."""
+delivery is enabled."""
 
 _MAX_EXACTLY_ONCE_DELIVERY_ACK_MODACK_RETRY_DURATION_SECS = 10 * 60
 """The maximum amount of time in seconds to retry failed acks and modacks when
-exactly-once is enabled."""
+exactly-once delivery is enabled."""
 
 
 class Dispatcher(object):
--- a/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
+++ b/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
@@ -201,7 +201,7 @@ class Dispatcher(object):
                 maximum=_MAX_EXACTLY_ONCE_ACK_MODACK_RETRY_DURATION_SECS,
             )
             while requests_to_retry:
-                time_to_wait = retry_delay_gen()
+                time_to_wait = next(retry_delay_gen)
                 _LOGGER.debug(
                     "Retrying {len(requests_to_retry)} ack(s) after delay of "
                     + str(time_to_wait)
@@ -288,7 +288,7 @@ class Dispatcher(object):
                 maximum=_MAX_EXACTLY_ONCE_ACK_MODACK_RETRY_DURATION_SECS,
             )
             while requests_to_retry:
-                time_to_wait = retry_delay_gen()
+                time_to_wait = next(retry_delay_gen)
                 _LOGGER.debug(
                     "Retrying {len(requests_to_retry)} modack(s) after delay of "
                     + str(time_to_wait)
@@ -299,7 +299,7 @@ class Dispatcher(object):
                 future_reqs_dict = {
                     req.ack_id: req for req in requests_to_retry if req.future
                 }
-                requests_to_retry = self._manager.send_unary_modack(
+                requests_completed, requests_to_retry = self._manager.send_unary_modack(
                     modify_deadline_ack_ids=[req.ack_id for req in requests_to_retry],
                     modify_deadline_seconds=[req.seconds for req in requests_to_retry],
                     future_reqs_dict=future_reqs_dict,
--- a/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
+++ b/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
@@ -22,6 +22,7 @@ import threading
 import typing
 from typing import List, Optional, Sequence, Union
 import warnings
+from google.api_core.retry import exponential_sleep_generator
 
 from google.cloud.pubsub_v1.subscriber._protocol import helper_threads
 from google.cloud.pubsub_v1.subscriber._protocol import requests
@@ -66,6 +67,13 @@ Accounting for some overhead, we should thus only send a maximum of 2500 ACK
 IDs at a time.
 """
 
+_MIN_ACK_MODACK_RETRY_DURATION_SECS = 1
+"""The time to wait for the first retry of failed acks and modacks when exactly-once
+is enabled."""
+
+_MAX_ACK_MODACK_RETRY_DURATION_SECS = 20 * 60
+"""The maximum amount of time in seconds to retry failed acks and modacks when
+exactly-once is enabled."""
 
 class Dispatcher(object):
     def __init__(self, manager: "StreamingPullManager", queue: "queue.Queue"):
@@ -168,17 +176,35 @@ class Dispatcher(object):
 
         # We must potentially split the request into multiple smaller requests
         # to avoid the server-side max request size limit.
-        ack_ids = (item.ack_id for item in items)
+        items_gen = iter(items)
+        ack_ids_gen = (item.ack_id for item in items)
         total_chunks = int(math.ceil(len(items) / _ACK_IDS_BATCH_SIZE))
 
         for _ in range(total_chunks):
-            request = gapic_types.StreamingPullRequest(
-                ack_ids=itertools.islice(ack_ids, _ACK_IDS_BATCH_SIZE)
-            )
-            self._manager.send(request)
+            future_reqs_dict = {req.ack_id: req for req in itertools.islice(items_gen, _ACK_IDS_BATCH_SIZE) if req.future}
+            requests_completed, requests_to_retry =  self._manager.send_unary_ack(
+                ack_ids=list(itertools.islice(ack_ids_gen, _ACK_IDS_BATCH_SIZE)),
+                future_reqs_dict=future_reqs_dict)
 
-        # Remove the message from lease management.
-        self.drop(items)
+            # Remove the completed messages from lease management.
+            self.drop(requests_completed)
+
+            # retry acks
+            retry_delay_gen = exponential_sleep_generator(initial=_MIN_ACK_MODACK_RETRY_DURATION_SECS,
+                                                          maximum=_MAX_ACK_MODACK_RETRY_DURATION_SECS)
+            while requests_to_retry:
+                time_to_wait = retry_delay_gen()
+                _LOGGER.debug("Retrying {len(requests_to_retry)} ack(s) after delay of " +
+                              str(time_to_wait) + " seconds")
+                time.sleep(time_to_wait)
+
+                future_reqs_dict = {req.ack_id: req for req in requests_to_retry if req.future}
+                requests_completed, requests_to_retry = self._manager.send_unary_ack(
+                    ack_ids=[req.ack_id for req in requests_to_retry],
+                    future_reqs_dict=future_reqs_dict)
+                assert len(requests_to_retry) <= _ACK_IDS_BATCH_SIZE, "Too many requests to be retried."
+                # Remove the completed messages from lease management.
+                self.drop(requests_completed)
 
     def drop(
         self,
@@ -215,16 +241,35 @@ class Dispatcher(object):
         """
         # We must potentially split the request into multiple smaller requests
         # to avoid the server-side max request size limit.
-        ack_ids = (item.ack_id for item in items)
-        seconds = (item.seconds for item in items)
+        items_gen = iter(items)
+        ack_ids_gen = (item.ack_id for item in items)
+        deadline_seconds_gen = (item.seconds for item in items)
         total_chunks = int(math.ceil(len(items) / _ACK_IDS_BATCH_SIZE))
 
         for _ in range(total_chunks):
-            request = gapic_types.StreamingPullRequest(
-                modify_deadline_ack_ids=itertools.islice(ack_ids, _ACK_IDS_BATCH_SIZE),
-                modify_deadline_seconds=itertools.islice(seconds, _ACK_IDS_BATCH_SIZE),
-            )
-            self._manager.send(request)
+            future_reqs_dict = {req.ack_id: req for req in itertools.islice(items_gen, _ACK_IDS_BATCH_SIZE) if req.future}
+            # no further work needs to be done for `requests_to_retry`
+            requests_completed, requests_to_retry = self._manager.send_unary_modack(
+                modify_deadline_ack_ids=list(itertools.islice(ack_ids_gen, _ACK_IDS_BATCH_SIZE)),
+                modify_deadline_seconds=list(itertools.islice(deadline_seconds_gen, _ACK_IDS_BATCH_SIZE)),
+                future_reqs_dict=future_reqs_dict)
+            assert len(requests_to_retry) <= _ACK_IDS_BATCH_SIZE, "Too many requests to be retried."
+
+            # retry modacks
+            retry_delay_gen = exponential_sleep_generator(initial=_MIN_ACK_MODACK_RETRY_DURATION_SECS,
+                                                          maximum=_MAX_ACK_MODACK_RETRY_DURATION_SECS)
+            while requests_to_retry:
+                time_to_wait = retry_delay_gen()
+                _LOGGER.debug("Retrying {len(requests_to_retry)} modack(s) after delay of " +
+                              str(time_to_wait) + " seconds")
+                time.sleep(time_to_wait)
+
+                print(requests_to_retry)
+                future_reqs_dict = {req.ack_id: req for req in requests_to_retry if req.future}
+                requests_to_retry = self._manager.send_unary_modack(
+                    modify_deadline_ack_ids=[req.ack_id for req in requests_to_retry],
+                    modify_deadline_seconds=[req.seconds for req in requests_to_retry],
+                    future_reqs_dict=future_reqs_dict)
 
     def nack(self, items: Sequence[requests.NackRequest]) -> None:
         """Explicitly deny receipt of messages.
@@ -233,6 +278,6 @@ class Dispatcher(object):
             items: The items to deny.
         """
         self.modify_ack_deadline(
-            [requests.ModAckRequest(ack_id=item.ack_id, seconds=0) for item in items]
+            [requests.ModAckRequest(ack_id=item.ack_id, seconds=0, future=item.future) for item in items]
         )
-        self.drop([requests.DropRequest(*item) for item in items])
+        self.drop([requests.DropRequest(ack_id=item.ack_id, byte_size=item.byte_size, ordering_key=item.ordering_key) for item in items])
--- a/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
+++ b/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
@@ -68,11 +68,11 @@ Accounting for some overhead, we should thus only send a maximum of 2500 ACK
 IDs at a time.
 """
 
-_MIN_EXACTLY_ONCE_ACK_MODACK_RETRY_DURATION_SECS = 1
+_MIN_EXACTLY_ONCE_DELIVERY_ACK_MODACK_RETRY_DURATION_SECS = 1
 """The time to wait for the first retry of failed acks and modacks when exactly-once
 is enabled."""
 
-_MAX_EXACTLY_ONCE_ACK_MODACK_RETRY_DURATION_SECS = 10 * 60
+_MAX_EXACTLY_ONCE_DELIVERY_ACK_MODACK_RETRY_DURATION_SECS = 10 * 60
 """The maximum amount of time in seconds to retry failed acks and modacks when
 exactly-once is enabled."""
 
@@ -217,8 +217,8 @@ class Dispatcher(object):
 
     def _retry_acks(self, requests_to_retry):
         retry_delay_gen = exponential_sleep_generator(
-            initial=_MIN_EXACTLY_ONCE_ACK_MODACK_RETRY_DURATION_SECS,
-            maximum=_MAX_EXACTLY_ONCE_ACK_MODACK_RETRY_DURATION_SECS,
+            initial=_MIN_EXACTLY_ONCE_DELIVERY_ACK_MODACK_RETRY_DURATION_SECS,
+            maximum=_MAX_EXACTLY_ONCE_DELIVERY_ACK_MODACK_RETRY_DURATION_SECS,
         )
         while requests_to_retry:
             time_to_wait = next(retry_delay_gen)
@@ -312,8 +312,8 @@ class Dispatcher(object):
 
     def _retry_modacks(self, requests_to_retry):
         retry_delay_gen = exponential_sleep_generator(
-            initial=_MIN_EXACTLY_ONCE_ACK_MODACK_RETRY_DURATION_SECS,
-            maximum=_MAX_EXACTLY_ONCE_ACK_MODACK_RETRY_DURATION_SECS,
+            initial=_MIN_EXACTLY_ONCE_DELIVERY_ACK_MODACK_RETRY_DURATION_SECS,
+            maximum=_MAX_EXACTLY_ONCE_DELIVERY_ACK_MODACK_RETRY_DURATION_SECS,
         )
         while requests_to_retry:
             time_to_wait = next(retry_delay_gen)
--- a/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
+++ b/google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
@@ -75,6 +75,7 @@ _MAX_EXACTLY_ONCE_ACK_MODACK_RETRY_DURATION_SECS = 10 * 60
 """The maximum amount of time in seconds to retry failed acks and modacks when
 exactly-once is enabled."""
 
+
 class Dispatcher(object):
     def __init__(self, manager: "StreamingPullManager", queue: "queue.Queue"):
         self._manager = manager
@@ -181,28 +182,43 @@ class Dispatcher(object):
         total_chunks = int(math.ceil(len(items) / _ACK_IDS_BATCH_SIZE))
 
         for _ in range(total_chunks):
-            future_reqs_dict = {req.ack_id: req for req in itertools.islice(items_gen, _ACK_IDS_BATCH_SIZE) if req.future}
-            requests_completed, requests_to_retry =  self._manager.send_unary_ack(
+            future_reqs_dict = {
+                req.ack_id: req
+                for req in itertools.islice(items_gen, _ACK_IDS_BATCH_SIZE)
+                if req.future
+            }
+            requests_completed, requests_to_retry = self._manager.send_unary_ack(
                 ack_ids=list(itertools.islice(ack_ids_gen, _ACK_IDS_BATCH_SIZE)),
-                future_reqs_dict=future_reqs_dict)
+                future_reqs_dict=future_reqs_dict,
+            )
 
             # Remove the completed messages from lease management.
             self.drop(requests_completed)
 
             # retry acks
-            retry_delay_gen = exponential_sleep_generator(initial=_MIN_EXACTLY_ONCE_ACK_MODACK_RETRY_DURATION_SECS,
-                                                          maximum=_MAX_EXACTLY_ONCE_ACK_MODACK_RETRY_DURATION_SECS)
+            retry_delay_gen = exponential_sleep_generator(
+                initial=_MIN_EXACTLY_ONCE_ACK_MODACK_RETRY_DURATION_SECS,
+                maximum=_MAX_EXACTLY_ONCE_ACK_MODACK_RETRY_DURATION_SECS,
+            )
             while requests_to_retry:
                 time_to_wait = retry_delay_gen()
-                _LOGGER.debug("Retrying {len(requests_to_retry)} ack(s) after delay of " +
-                              str(time_to_wait) + " seconds")
+                _LOGGER.debug(
+                    "Retrying {len(requests_to_retry)} ack(s) after delay of "
+                    + str(time_to_wait)
+                    + " seconds"
+                )
                 time.sleep(time_to_wait)
 
-                future_reqs_dict = {req.ack_id: req for req in requests_to_retry if req.future}
+                future_reqs_dict = {
+                    req.ack_id: req for req in requests_to_retry if req.future
+                }
                 requests_completed, requests_to_retry = self._manager.send_unary_ack(
                     ack_ids=[req.ack_id for req in requests_to_retry],
-                    future_reqs_dict=future_reqs_dict)
-                assert len(requests_to_retry) <= _ACK_IDS_BATCH_SIZE, "Too many requests to be retried."
+                    future_reqs_dict=future_reqs_dict,
+                )
+                assert (
+                    len(requests_to_retry) <= _ACK_IDS_BATCH_SIZE
+                ), "Too many requests to be retried."
                 # Remove the completed messages from lease management.
                 self.drop(requests_completed)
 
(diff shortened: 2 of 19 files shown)


google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -168,17 +168,30 @@ def ack(self, items: Sequence[requests.AckRequest]) -> None:
...
+            requests_completed, requests_to_retry =  self._manager.send_unary_ack(
+                ack_ids=list(itertools.islice(ack_ids_gen, _ACK_IDS_BATCH_SIZE)),
+                future_reqs_dict=future_reqs_dict)
+
+            # Remove the completed messages from lease management.
+            self.drop(requests_completed)
+
+            # retry acks
+            # todo: figure out if retry delays/limits are needed
+            while requests_to_retry:
```
plamut, line 186: How long can this loop spin around?

If there is a transient error, a failed unary ACK request could be sent many times in a quick succession here, which is undesirable.

In addition, the `ack()` method could block for a long time while eagerly retrying, effectively blocking the `dispatch_callback()` method and the entire dispatcher. This means, for instance, that any new lease requests added to the dispatch queue will not be processed in time and leased messages could unnecessarily expire.

If we really want to repeatedly make requests here, it should be delegated to a separate thread.

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -168,17 +168,30 @@ def ack(self, items: Sequence[requests.AckRequest]) -> None:
...
+                _LOGGER.debug("Retrying " + str(len(requests_to_retry)) + " acks")
+                future_reqs_dict = {req.ack_id: req for req in requests_to_retry if req.future}
+                requests_completed, requests_to_retry = self._manager.send_unary_ack(
+                    ack_ids=[req.ack_id for req in requests_to_retry],
+                    future_reqs_dict=future_reqs_dict)
+                assert len(requests_to_retry) <= _ACK_IDS_BATCH_SIZE, "Too many requests to be retried."
```
plamut, line 192: Is this meant to be used as a runtime check in production, i.e. not just when running tests?

If so, an exception should be explicitly raised if the invariant does not hold. The problem with `assert` statements is that they are removed if Python is run in optimized mode, demo:
```
$ python -c "x = 0; assert x != 0; print(1/x)"
  File "<string>", line 1, in <module>
AssertionError

$ python -O -c "x = 0; assert x != 0; print(1/x)"  # optimized mode
  File "<string>", line 1, in <module>
ZeroDivisionError: division by zero

```

plamut (commented):
This is only a partial feedback, I have not gone through the entire PR yet.

The think we need to be careful about is to not block any worker threads for too long, e.g. when retrying failed requests.

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
plamut: Also applies to the `drop()` method, which is very similar.

(we could also consider extracting some of the common logic out, if it does not hinder readability)

google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -242,9 +243,44 @@ def ack(self) -> None:
                 byte_size=self.size,
                 time_to_ack=time_to_ack,
                 ordering_key=self.ordering_key,
+                future=None
             )
         )
 
+    def ack_with_response(self) -> None:
+        """Acknowledge the given message.
```
plamut, line 251: The only difference compared to `ack()` is whether or not we use and return a future. It would be IMO better for maintenance if we instead just add an optional `with_response` argument to `ack()` (`False` by default) and set the value of `future` accordingly.

google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -281,16 +317,70 @@ def modify_ack_deadline(self, seconds: int) -> None:
                 against.
         """
         self._request_queue.put(
-            requests.ModAckRequest(ack_id=self._ack_id, seconds=seconds)
+            requests.ModAckRequest(ack_id=self._ack_id, seconds=seconds, future=None)
         )
 
+    def modify_ack_deadline_with_response(self, seconds: int) -> "pubsub_v1.subscriber.futures.Future":
+        """Resets the deadline for acknowledgement and returns the response
+        status via a future.
```
plamut, line 325: Same here, we can de-ducplicate the code by introducing an optional `with_response` kwarg.

google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -281,16 +317,70 @@ def modify_ack_deadline(self, seconds: int) -> None:
...
+        return future
+
     def nack(self) -> None:
-        """Decline to acknowldge the given message.
+        """Decline to acknowledge the given message.
 
-        This will cause the message to be re-delivered to the subscription.
+        This will cause the message to be re-delivered to subscribers. Re-deliveries
+        may take place immediately or after a delay, and may arrive at this subscriber
+        or another.
```
plamut, line 356: I see this extra info can be very useful to developers. :+1:

google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -281,16 +317,70 @@ def modify_ack_deadline(self, seconds: int) -> None:
...
             requests.NackRequest(
-                ack_id=self._ack_id, byte_size=self.size, ordering_key=self.ordering_key
+                ack_id=self._ack_id, byte_size=self.size, ordering_key=self.ordering_key,
+                future=None
+            )
+        )
+
+    def nack_with_response(self) -> None:
+        """Decline to acknowledge the given message, returning the response status via
+        a future.
```
plamut, line 367: Same here, let's just add an optional argument `with_response` .

google/pubsub_v1/types/pubsub.py
```diff
@@ -1114,11 +1114,14 @@ class SubscriptionProperties(proto.Message):
...
 
         Attributes:
+            exactly_once_delivery_enabled (bool):
+
             message_ordering_enabled (bool):
                 True iff message ordering is enabled for this
                 subscription.
         """
 
+        exactly_once_delivery_enabled = proto.Field( proto.BOOL, number=1,)
```
plamut, line 1124: Related to the first comment in the other review, should we persist this change through `owlbot.py` or are we expecting the proto change to go public before we merge this?

tests/unit/pubsub_v1/subscriber/test_dispatcher.py
```diff
@@ -259,21 +259,22 @@ def test_modify_ack_deadline_splitting_large_payload():
...
     all_ack_ids = {item.ack_id for item in items}
     sent_ack_ids = collections.Counter()
 
     for call in calls:
-        message = call.args[0]
-        assert message._pb.ByteSize() <= 524288  # server-side limit (2**19)
-        sent_ack_ids.update(message.modify_deadline_ack_ids)
+        modack_ackids = call[1]["modify_deadline_ack_ids"]
+        assert len(modack_ackids) <= dispatcher._ACK_IDS_BATCH_SIZE
+        sent_ack_ids.update(modack_ackids)
```
plamut, line 277: This seems to have changed the nature of the test, is that a find/replace error?

We should be checking the message byte size here.

plamut (changes requested):
Round two, added some additional comments.

Will review the streaming pull manager tests later.

tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
```diff
@@ -860,9 +956,72 @@ def test__on_response_modifies_ack_deadline():
...
+
+        # mod-acks called with 60 sec min lease value for exactly_once subscriptions
+        dispatcher.modify_ack_deadline([requests.ModAckRequest("ack_3", 60, None), requests.ModAckRequest("ack_4", 60, None)])
     )
 
+    # exactly_once is still disabled b/c subscription_properties says so
+    manager._on_response(response1)
+
+    # exactly_once should be enabled after this request b/c subscription_properties says so
+    manager._on_response(response2)
```
plamut, line 1023: It's not entirely clear what these two calls do, and there's no visible assertion. A comment explaining this would be useful.

tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
```diff
@@ -1130,3 +1357,176 @@ def test_activate_ordering_keys_stopped_scheduler():
...
+    # Unpack() failed
+    unpack.return_value = None
+
+    assert not streaming_pull_manager._get_ack_errors(exception)
+
+@mock.patch("grpc_status.rpc_status.from_call")
+def test_get_ack_errors(from_call):
+    manager = make_manager()
+
+    # exception = exceptions.from_grpc_status(grpc.StatusCode.INTERNAL, "message")
```
plamut, line 1385: Forgotten comment, should be deleted?

tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
```diff
@@ -1130,3 +1357,176 @@ def test_activate_ordering_keys_stopped_scheduler():
...
+    from_call.side_effect = None
+    from_call.return_value = st
+    # happy case - errors returned in a map
+    ack_errors = streaming_pull_manager._get_ack_errors(exception)
+    assert ack_errors
+    assert ack_errors["ack_1"] == "error1"
+
+
+def test_process_errors():
+    # no requests so no items in results lists
```
plamut, line 1425: This unit test tests many different aspects of `_process_errors()`, let's split it into specialized test cases.

(the same can probably be done with `test_get_ack_errors()`)

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -104,6 +108,61 @@ def _wrap_callback_errors(
...
+    except ValueError:
+        _LOGGER.debug( "ValueError when parsing ErrorInfo.", exc_info=True)
+        return None
+
+def _process_errors(future_reqs_dict: "containers.ScalarMap",
+                    errors_dict: Optional["containers.ScalarMap"]):
+    """Process errors returned by the server in `errors_dict` by completing
+    the request futures in `future_reqs_dict` (with a success or exception) or
+    returning requests for further retries.
+    """
```
plamut, line 139: (nit) This docstring formatting does not follow the common form and is likely to render strangely in the docs.

plamut (commented):
Went through the rest of the PR, this should be it for the first review round. :)

google/cloud/pubsub_v1/proto/pubsub.proto
```diff
@@ -1164,6 +1164,7 @@ message StreamingPullRequest {
 message StreamingPullResponse {
   // Subscription properties sent as part of the response.
   message SubscriptionProperties {
+    bool exactly_once_delivery_enabled = 1;
```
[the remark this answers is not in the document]
pradn: This change will be checked in after the proto change has been made public and the autogen has been checked in to this repo. I'll rebase this change on top of that update. So, we'll be fine I think. I don't anticipate any big changes aside from your type hint change, so I hope we don't have too much more churn with this CL.

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -168,17 +168,30 @@ def ack(self, items: Sequence[requests.AckRequest]) -> None:
 
         # We must potentially split the request into multiple smaller requests
         # to avoid the server-side max request size limit.
-        ack_ids = (item.ack_id for item in items)
+        items_gen = (i for i in items)
```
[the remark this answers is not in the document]
pradn: Thanks, didn't know this!

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
pradn: Good to know! I intend it to be closer to invariant documentation. I'm not relying on it to run in prod.

google/pubsub_v1/types/pubsub.py
pradn: Same answer as another. We expect the proto change to precede this change.

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
pradn: What's the conventional format? Are there some docs I can read about this?

plamut: One way is to check the existing methods, but I also found [this guide](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods).

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -435,48 +515,73 @@ def _schedule_message_on_hold(
...
+                "error resolution for too long, will now trigger shutdown.",
+                exc_info=False,
             )
+            # The underlying channel has been suffering from a retryable error
+            # for too long, time to give up and shut the streaming pull down.
+            self._on_rpc_done(exc)
+            raise
 
-        if request.modify_deadline_ack_ids:
+        requests_completed, requests_to_retry = _process_errors(future_reqs_dict, ack_errors_dict)
```
plamut, line 551: BTW, the name `_process_errors()` is a bit misleading, since we do not actually process any errors if the `acknowledge()` call succeeds.

plamut:
@pradn I see a lot of comments marked as resolved, but no changes in the code - is there a `git push` missing? :)

pradn (commented):
Pushed

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
plamut: `_process_futures()` is a good choice IMO. :+1:

google/cloud/pubsub_v1/subscriber/message.py
plamut: I agree with 3 or more, while two instances of the same are sometimes borderline. I would personally still go with de-duplication here, but I also understand a different view.

tests/unit/pubsub_v1/subscriber/test_dispatcher.py
plamut: True, merely summing the individual strings would be a bit off. But what if we actually build a `ModifyAckDeadlineRequest` instance in the test and check the size of that? Since that's the payload that is sent to the backend?

tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py
plamut: Sorry, I might be missing something, which comments were added? What is the role of `manager._on_response(...)` calls here, what errors/assertions do they trigger?

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
plamut: @pradn This seems to have slipped through, are we sure this will behave gracefully in all cases? Won't it block the thread for too long?

pradn: I plan to fix this. I'd like to get some initial testing done before. (Should have mentioned this to you :) )

plamut: OK good then. :+1: For a moment I got worried that we forgot about this. :)

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -66,6 +67,13 @@
 IDs at a time.
 """
 
+_MIN_ACK_MODACK_RETRY_DURATION_SECS = 1
+"""The time to wait for the first retry of failed acks and modacks when exactly-once
+is enabled."""
+
+_MAX_ACK_MODACK_RETRY_DURATION_SECS = 20 * 60
```
maheshgattani, line 74: This is 20 mins. The max ack deadline is 10 mins.  What's the reason to make this 20?

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -66,6 +67,13 @@
 IDs at a time.
 """
 
+_MIN_ACK_MODACK_RETRY_DURATION_SECS = 1
+"""The time to wait for the first retry of failed acks and modacks when exactly-once
```
maheshgattani, line 71: Should the variable name contain "exactly_once"

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -168,17 +176,35 @@ def ack(self, items: Sequence[requests.AckRequest]) -> None:
...
+            future_reqs_dict = {req.ack_id: req for req in itertools.islice(items_gen, _ACK_IDS_BATCH_SIZE) if req.future}
+            requests_completed, requests_to_retry =  self._manager.send_unary_ack(
+                ack_ids=list(itertools.islice(ack_ids_gen, _ACK_IDS_BATCH_SIZE)),
+                future_reqs_dict=future_reqs_dict)
+
+            # Remove the completed messages from lease management.
+            self.drop(requests_completed)
+
+            # retry acks
+            retry_delay_gen = exponential_sleep_generator(initial=_MIN_ACK_MODACK_RETRY_DURATION_SECS,
```
maheshgattani, line 193: This seems to be happening with or without exactly once? This should happen only for exactly once subs. Am I understanding this correctly?

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -215,16 +241,35 @@ def modify_ack_deadline(self, items: Sequence[requests.ModAckRequest]) -> None:
...
+            future_reqs_dict = {req.ack_id: req for req in itertools.islice(items_gen, _ACK_IDS_BATCH_SIZE) if req.future}
+            # no further work needs to be done for `requests_to_retry`
+            requests_completed, requests_to_retry = self._manager.send_unary_modack(
+                modify_deadline_ack_ids=list(itertools.islice(ack_ids_gen, _ACK_IDS_BATCH_SIZE)),
+                modify_deadline_seconds=list(itertools.islice(deadline_seconds_gen, _ACK_IDS_BATCH_SIZE)),
+                future_reqs_dict=future_reqs_dict)
+            assert len(requests_to_retry) <= _ACK_IDS_BATCH_SIZE, "Too many requests to be retried."
+
+            # retry modacks
+            retry_delay_gen = exponential_sleep_generator(initial=_MIN_ACK_MODACK_RETRY_DURATION_SECS,
```
maheshgattani, line 259: Same comment as above: is this happening for all subs or just exactly once?

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -104,6 +112,63 @@ def _wrap_callback_errors(
...
+    the request futures in `future_reqs_dict` (with a success or exception) or
+    to return requests for further retries.
+    """
+    requests_completed = []
+    requests_to_retry = []
+    for ack_id in future_reqs_dict:
+        if errors_dict and ack_id in errors_dict:
+            exactly_once_error = errors_dict[ack_id]
+            if exactly_once_error.startswith("PERMANENT_"):
+                exc = RuntimeError("Permanent error: " + exactly_once_error)
```
maheshgattani, line 152: I am not sure I understand how this exception is handled. Can you explain? Making sure this will not crash the program.

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -273,6 +347,20 @@ def _obtain_ack_deadline(self, maybe_update: bool) -> float:
                     histogram.MIN_ACK_DEADLINE,
                 )
                 self._ack_deadline = min(self._ack_deadline, flow_control_setting)
+
+            # If the user explicitly sets a min ack_deadline, respect it.
+            if self.flow_control.min_duration_per_lease_extension > 0:
+                # The setting in flow control could be too high, adjust if needed.
+                flow_control_setting = min(
```
maheshgattani, line 354: If we see streaming_pull_response.enable_exactly_once_delivery as true and the user hasn't provided a custom min duration, we should update the min duration to 60 seconds. I do not see this happening in this PR? If I am missing it, please point me to where this is happening.

google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -242,9 +243,44 @@ def ack(self) -> None:
                 byte_size=self.size,
                 time_to_ack=time_to_ack,
                 ordering_key=self.ordering_key,
+                future=None
             )
         )
 
+    def ack_with_response(self) -> "pubsub_v1.subscriber.futures.Future":
```
maheshgattani, line 250: As discussed offline, can we add a code sample using:

- ack_with_response
- nack_with_response
- modify_ack_deadline_with_response
- min_duration_per_lease_extension

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -733,11 +863,20 @@ def _on_response(self, response: gapic_types.StreamingPullResponse) -> None:
             self._on_hold_bytes,
         )
 
+        with self._exactly_once_enabled_lock:
+            if response.subscription_properties.exactly_once_delivery_enabled != \
+               self._exactly_once_enabled:
+                 self._exactly_once_enabled = response.subscription_properties.exactly_once_delivery_enabled
+                 # Update ack_deadline, whose minimum depends on self._exactly_once_enabled
+                 # This method acquires the self._ack_deadline_lock lock.
+                 self._obtain_ack_deadline(maybe_update=True)
```
pradn, line 872: Mahesh, look here for the logic to update the min ack deadline when we see `subscription_properties.exactly_once_delivery_enabled` has changed. I also have a test for this.

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
maheshgattani: Resolving as per another comment.

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
pradn: This code path is only executed if there's futures with errors, and that only happens if 1) the user invokes a ack/modack_with_future method 2) the server comes back with an error for that ack id. Both of these only occur when the user is using exactly-once. Note we don't check if exactly-once is enabled at the stream level for this behavior. I think it's wise not to b/c we'll better handle edge cases around enablement/disablement. For example, if the server sends an ack error and we had just learned that exactly-once has been disabled at the stream level, we still want to tell the user the ack failed.

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
pradn: The future is completed with an exception. If the user calls the future's result() method, it'll throw an exception. They should check if the future is done successfully first. This is standard behavior for futures in Python.

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
pradn: I meant ack/modack_with_result, not *with_future.

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
pradn: Just exactly-once subs - same explanation.

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
pradn: It's a just-in-case thing. I will change it to 10min since there's no stronger reason than that.

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
pradn: Changed.

tests/unit/pubsub_v1/subscriber/test_futures_subscriber.py
```diff
@@ -76,3 +80,30 @@ def test_cancel(self):
...
+        future = futures.Future()
+        future.set_exception(
+            AcknowledgeError(
+                AcknowledgeStatus.PERMISSION_DENIED, "Something bad happened."
+            )
+        )
+        with pytest.raises(AcknowledgeError) as e:
+            future.result()
+            assert e.error_code == AcknowledgeStatus.PERMISSION_DENIED
+            assert e.info == "Something bad happened."
```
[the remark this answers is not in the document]
pradn: Thank you for your help!

acocuzzo (approved):
After conversation: https://github.com/googleapis/python-pubsub/pull/550/files#r769774809
 is resolved, and all checks pass, LGTM.

samples/snippets/subscriber.py
```diff
@@ -580,6 +580,57 @@ def callback(message: pubsub_v1.subscriber.message.Message) -> None:
...
+            # When `timeout` is not set, result() will block indefinitely,
+            # unless an exception is encountered first.
+            streaming_pull_future.result(timeout=timeout)
+        except TimeoutError:
+            streaming_pull_future.cancel()  # Trigger the shutdown.
+            streaming_pull_future.result()  # Block until the shutdown is complete.
+    # [END pubsub_subscriber_async_pull]
+    # [END pubsub_quickstart_subscriber]
+
+
```
anguillanneuf, line 633: @pradn We need to remove this sample because it uses identical region tags as other snippets in the file. It will confuse our devsite docs generation.

anguillanneuf: Thanks @pradn. I see `ack_with_response()` is different. Let me go and create these region tags first and let's update these to match.

samples/snippets/subscriber.py
```diff
@@ -580,6 +580,57 @@ def callback(message: pubsub_v1.subscriber.message.Message) -> None:
...
+    subscriber = pubsub_v1.SubscriberClient()
+    # The `subscription_path` method creates a fully qualified identifier
+    # in the form `projects/{project_id}/subscriptions/{subscription_id}`
+    subscription_path = subscriber.subscription_path(project_id, subscription_id)
+
+    def callback(message: pubsub_v1.subscriber.message.Message) -> None:
+        print(f"Received {message}.")
+
+        ack_future = message.ack_with_response()
```
anguillanneuf, line 607: Please add a comment for what `ack_with_response` does.

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
pradn: Fixed: https://github.com/googleapis/python-pubsub/pull/550/commits/d7305e292defc509762f163c7ed2ea4ac87a73d0

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -435,48 +552,78 @@ def _schedule_message_on_hold(
...
+        except exceptions.GoogleAPICallError as exc:
+            _LOGGER.debug(
+                "Exception while sending unary RPC. This is typically "
+                "non-fatal as stream requests are best-effort.",
+                exc_info=True,
             )
+            error_status = _get_status(exc)
+            ack_errors_dict = _get_ack_errors(exc)
+        except exceptions.RetryError as exc:
+            _LOGGER.debug(
```
maheshgattani, line 578: The comment concerns be: what is shutting down?

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -700,6 +867,36 @@ def _get_initial_request(
...
+                future = futures.Future()
+                request = requests.ModAckRequest(ack_id, ack_deadline, future)
+                items.append(request)
+
+            assert self._dispatcher is not None
+            self._dispatcher.modify_ack_deadline(items)
+
+            for req in items:
+                try:
+                    req.future.result()
```
maheshgattani, line 886: This will block right? Any concerns around performance?

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -700,6 +867,36 @@ def _get_initial_request(
...
+                except AcknowledgeError:
+                    _LOGGER.warning(
+                        "AcknowledgeError when modacking a message immediately after receiving it.",
```
maheshgattani, line 889: Is this comment correct? Looks like this code will be executed more often than just after receiving the message (and for continuous lease extension)

google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -242,9 +243,43 @@ def ack(self) -> None:
...
+        """Acknowledge the given message.
+
+        Acknowledging a message in Pub/Sub means that you are done
+        with it, and it will not be delivered to this subscription again.
+        You should avoid acknowledging messages until you have
+        *finished* processing them, so that in the event of a failure,
+        you receive the message again.
+
+        .. warning::
+            Acks in Pub/Sub are best effort. You should always
```
maheshgattani, line 260: With exactly once this isn't correct. Please add what happens when users use this method with and without exactly once.

google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -281,16 +316,74 @@ def modify_ack_deadline(self, seconds: int) -> None:
                 against.
         """
         self._request_queue.put(
-            requests.ModAckRequest(ack_id=self._ack_id, seconds=seconds)
+            requests.ModAckRequest(ack_id=self._ack_id, seconds=seconds, future=None)
```
maheshgattani, line 319: What effect does future=None do here? Can you add a comment please?

google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -281,16 +316,74 @@ def modify_ack_deadline(self, seconds: int) -> None:
...
+        you should not need to manually deal with setting ack deadlines. The exception case is
+        if you are implementing your own custom subclass of
+        :class:`~.pubsub_v1.subcriber._consumer.Consumer`.
+
+        Args:
+            seconds:
+                The number of seconds to set the lease deadline to. This should be
+                between 0 and 600. Due to network latency, values below 10 are advised
+                against.
+        Returns:
```
maheshgattani, line 338: Please document what to expect from the future and how to interpret it.

google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -242,9 +243,43 @@ def ack(self) -> None:
...
+        *finished* processing them, so that in the event of a failure,
+        you receive the message again.
+
+        .. warning::
+            Acks in Pub/Sub are best effort. You should always
+            ensure that your processing code is idempotent, as you may
+            receive any given message more than once.
+
+        Returns:
+            A :class:`~google.cloud.pubsub_v1.subscriber.futures.Future`
```
maheshgattani, line 265: Please add how to interpret the future

google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -281,16 +316,74 @@ def modify_ack_deadline(self, seconds: int) -> None:
...
+    def nack_with_response(self) -> "futures.Future":
+        """Decline to acknowledge the given message, returning the response status via
+        a future.
+
+        This will cause the message to be re-delivered to subscribers. Re-deliveries
+        may take place immediately or after a delay, and may arrive at this subscriber
+        or another.
+
+        Returns:
+            A :class:`~google.cloud.pubsub_v1.subscriber.futures.Future`
```
maheshgattani, line 375: Please add how to interpret this return value

google/cloud/pubsub_v1/subscriber/_protocol/requests.py
```diff
@@ -39,9 +44,11 @@ class LeaseRequest(NamedTuple):
 class ModAckRequest(NamedTuple):
     ack_id: str
     seconds: float
+    future: Optional["futures.Future"]
```
maheshgattani, line 47: Where is this implemented? Sorry I am struggling to find it in the CL.

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
pradn: My mistake, fixed.

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
pradn: Ack and modack RPCs calls have always been synchronous in this code path so this doesn't change that fact. So we're fine.

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
pradn: Grep'd and updated uses of "exactly-once" to include the "delivery" part.

google/cloud/pubsub_v1/subscriber/_protocol/requests.py
pradn: Whether a future is set or not depends on whether the user calls ack or ack_with_response (or the modack & nack equivalents). This is done in `google/cloud/pubsub_v1/subscriber/message.py`.

If a future is present, we put it in an ack_id->future in dispatcher.py and pass it to the streaming_pull_manager (send_unary_ack or send_unary_modack methods). Those methods are responsible for finishing the futures or returning failed ones to be retired by the dispatcher.

google/cloud/pubsub_v1/subscriber/message.py
pradn: Fixed.

google/cloud/pubsub_v1/subscriber/message.py
pradn: Fixed.

google/cloud/pubsub_v1/subscriber/message.py
pradn: Fixed.

tests/unit/pubsub_v1/subscriber/test_dispatcher.py
pradn: I think the complexity of that isn't worth it.

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -233,6 +312,20 @@ def nack(self, items: Sequence[requests.NackRequest]) -> None:
...
         self.modify_ack_deadline(
-            [requests.ModAckRequest(ack_id=item.ack_id, seconds=0) for item in items]
+            [
+                requests.ModAckRequest(
+                    ack_id=item.ack_id, seconds=0, future=item.future
+                )
+                for item in items
+            ]
+        )
+        self.drop(
```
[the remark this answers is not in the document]
pradn: self.modify_ack_deadlines retries failures as needed and eventually completes all of the futures. After that, we do need to drop the ack_ids from lease management.

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
pradn: It's the bidi stream itself. If no progress whatsoever can be made - no calls are succeeding - we shut down the stream. The user receives an error message for the entire subscribe process. ie: there's a subscribe future that users can use to track the entire subscribe process / stream itself, and that is failed. So users will have to recreate the subscriber client. We should leave this behavior as is.

I believe I'm missing the completion of the futures with a failure in this case.

I think we should leave this behavior as is.

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -66,6 +68,14 @@
 IDs at a time.
 """
 
+_MIN_EXACTLY_ONCE_DELIVERY_ACK_MODACK_RETRY_DURATION_SECS = 1
+"""The time to wait for the first retry of failed acks and modacks when exactly-once
```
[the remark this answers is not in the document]
pradn: Hmm, I think I made the changes and somehow lost them. Will fix.

maheshgattani (approved):

PR #551, opened 2021-12-17
process: add mypy session for samples

plamut:

This PR adds the `mypy` session for samples and cleans up their type annotations.

However, the PR does not address the errors caused by `mypy` not realizing that a lot of methods from the generated client are dynamically injected into the hand written class. It also does not deal with a few missing annotations in `google-api-core` (e.g. `retry.Retry`).

Additional thing to consider is that the client classes that the end users use now have many more methods on them that were previously stored on their `client.api` attribute (and the .`api` attribute has been removed). Caution is needed here.

### DIFF
--- a/google/cloud/pubsub_v1/publisher/futures.py
+++ b/google/cloud/pubsub_v1/publisher/futures.py
@@ -69,7 +69,7 @@ class Future(futures.Future):
     def add_done_callback(
         self, callback: Callable[["pubsub_v1.publisher.futures.Future"], Any]
     ) -> None:
-        """Attache a callable that will be called when the future finishes.
+        """Attach a callable that will be called when the future finishes.
 
         Args:
             callback:
--- a/google/cloud/pubsub_v1/publisher/futures.py
+++ b/google/cloud/pubsub_v1/publisher/futures.py
@@ -14,9 +14,13 @@
 
 from __future__ import absolute_import
 
-from typing import Union
+import typing
+from typing import Any, Callable, Union
 
 from google.cloud.pubsub_v1 import futures
+
+if typing.TYPE_CHECKING:  # pragma: NO COVER
+    from google.cloud import pubsub_v1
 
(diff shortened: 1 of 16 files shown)


plamut:
Hm... this will likely not be mergeable, because `mypy_samples` currently fails (as a part of the main Kokoro run). Probably best to temporarily disable that session.

google/cloud/pubsub_v1/publisher/futures.py
```diff
@@ -60,3 +64,20 @@ def result(self, timeout: Union[int, float] = None) -> str:
                 call execution.
         """
         return super().result(timeout=timeout)
+
+    # This exists to make the type checkers happy.
+    def add_done_callback(
+        self, callback: Callable[["pubsub_v1.publisher.futures.Future"], Any]
+    ) -> None:
+        """Attache a callable that will be called when the future finishes.
```
tswast, line 72: Nit: small typo

```suggestion
        """Attach a callable that will be called when the future finishes.
```

noxfile.py
```diff
@@ -47,6 +47,7 @@
     "blacken",
     "mypy",
     "pytype",
+    # "mypy_samples",  # TODO: uncomment when the checks pass
```
tswast, line 50: Is this intended to stay commented out?

If so, maybe link to https://github.com/googleapis/python-pubsub/issues/536

tswast (approved):
LGTM except for a couple of nits.

noxfile.py
tswast: Edit: Just saw the comment above. I'm okay leaving this as-is, assuming it's very temporary.

plamut: Yes, it's very temporary indeed. I wanted to keep the refactoring changes from #536 separate for easier rollback, should that be needed (so that we wouldn't be forced to purge other things as well).

PR #552, opened 2021-12-22
fix: refactor client classes for safer type checking

plamut:

This PR is based on https://github.com/googleapis/python-pubsub/pull/551, it addresses the false positives `mypy` produces due to the fact that it cannot know that most of the generated methods are dynamically injected into the hand-written client wrapper classes.

googleapis/python-pubsub#551 should likely be merged first, as this change is somewhat more risky, and we want to have it in its own commit to make the rollback easier and smaller, should that be necessary.


samples/snippets/requirements.txt
```diff
@@ -1,2 +1,6 @@
+# Version 2.1.1 of google-api-core version is the first type-checked release.
+google-api-core[grpc]==2.1.1
```
tseaver, line 2: Shouldn't this be a minimum pin?

samples/snippets/requirements.txt
```diff
@@ -1,2 +1,6 @@
...
+# Version 2.2.0 of google-cloud-core version is the first type-checked release.
+google-cloud-core==2.2.0
```
tseaver, line 4: And likewise here.

google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -206,12 +200,15 @@ def _get_or_create_sequencer(self, topic: str, ordering_key: str) -> SequencerTy
         sequencer_key = (topic, ordering_key)
         sequencer = self._sequencers.get(sequencer_key)
         if sequencer is None:
+            # Disable pytype false positive...
+            # pytype: disable=wrong-arg-types
```
tseaver, line 204: Can we just remove `pytype` altogether?

google/cloud/pubsub_v1/subscriber/client.py
```diff
@@ -243,6 +232,8 @@ def callback(message):
         """
         flow_control = types.FlowControl(*flow_control)
 
+        # Disable pytype false positive...
+        # pytype: disable=wrong-arg-types
```
tseaver, line 236: Likewise, removing `pytype` seems like the Right Thing(TM) here.

google/cloud/pubsub_v1/publisher/client.py
plamut: In fact, we can, and will do it. Can't really see much value compared to `mypy`.

I will make the changes in https://github.com/googleapis/python-pubsub/pull/551, since I'd like to keep the class refactoring part separate (for an easier rollback, should that be needed).

samples/snippets/requirements.txt
plamut: Hmm... I think this should actually be moved into the `mypy_samples` nox session, as the change is primarily meant for the type check's benefits.

Interestingly though, the samples tests already **don't** use the minimum version - they just [install stuff](https://github.com/googleapis/python-pubsub/blob/17b5d143f2ffcab7e7efea8bb2018148c00ce1ee/samples/snippets/noxfile.py#L191-L195) from `requirements.txt` (with no constraints on api-core), and by default `pip` installs the most recent compatible versions.

I'll check with the samples team on that.

plamut: Got a response, the samples are actually tested with the latest library release. Sounds like the version pin should just be updated to `>=` (or even be omitted altogether, but I prefer to be explicit here).

tswast (commented):
Looks a lot cleaner! A couple of questions.

google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -146,8 +140,8 @@ def __init__(
 
         # Add the metrics headers, and instantiate the underlying GAPIC
         # client.
-        self.api = publisher_client.PublisherClient(**kwargs)
```
tswast, line 149: Should we call out that this was removed and mark this as a breaking change?

I don't think this would have appeared in the API reference docs, so it's not quite a fully public property, but it is named like one.

google/cloud/pubsub_v1/subscriber/client.py
```diff
@@ -123,11 +117,6 @@ def target(self) -> str:
         """
         return self._target
 
-    @property
-    def api(self) -> subscriber_client.SubscriberClient:
-        """The underlying gapic API client."""
-        return self._api
-
```
tswast, line 130: Since this was documented

https://googleapis.dev/python/pubsub/latest/subscriber/api/client.html?highlight=api#google.cloud.pubsub_v1.subscriber.client.Client.api

we definitely need to consider this a breaking change. IMO it is worth taking that breaking change on, though.

Alternatively, you could update this to return `super()` or something.

tswast (changes requested):

google/cloud/pubsub_v1/publisher/client.py
plamut: Yes, we should indeed do that, since `api` is a public attribute.

google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -252,7 +246,12 @@ def resume_publish(self, topic: str, ordering_key: str) -> None:
             else:
                 sequencer.unpause()
 
-    def publish(
+    def _gapic_publish(self, *args, **kwargs) -> "pubsub_types.PublishResponse":
+        """TODO: docstrings"""
+        # Call GAPIC publish API directly
```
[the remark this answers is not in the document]
plamut: Thanks for catching that, that's precisely what the TODO was meant for. :+1: I'll add the docstrings.

google/cloud/pubsub_v1/subscriber/client.py
plamut: I agree, we should preserve this property.

I'm also thinking of deprecating it (docstring + a warning) now that the hand-written instances inherit all the essential stuff (including the gRPC channel, etc.).

noxfile.py
```diff
@@ -46,7 +44,7 @@
     "lint_setup_py",
     "blacken",
     "mypy",
-    "pytype",
+    # "mypy_samples",  # TODO: uncomment when the checks pass
```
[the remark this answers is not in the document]
plamut: Definitely, should have been enabled here already. Good catch.

google/cloud/pubsub_v1/subscriber/client.py
tswast: Adding it back with a deprecation sounds like the right plan to me

plamut:
@tswast Applied the suggestions.

Sorry for the force push, rebased just the relevant commits onto the latest `main` after #551 was merged.

tswast:
Getting a few mypy errors on the samples

```
nox > Running session mypy_samples
nox > Creating virtual environment (virtualenv) using python3.8 in .nox/mypy_samples
nox > python -m pip install -e '.[all]'
nox > python -m pip install pytest
nox > python -m pip install mypy==0.910
nox > python -m pip install types-mock types-protobuf types-setuptools
nox > mypy --config-file /tmpfs/src/github/python-pubsub/samples/snippets/mypy.ini --no-incremental samples/
samples/snippets/utilities/us_states_pb2.py:11: error: Call to untyped function "Default" in typed context
samples/snippets/subscriber.py:107: error: Module "google.cloud.pubsub_v1.types" has no attribute "DeadLetterPolicy"
samples/snippets/subscriber.py:172: error: Module has no attribute "PushConfig"
samples/snippets/subscriber.py:260: error: Module has no attribute "PushConfig"
samples/snippets/subscriber.py:262: error: Module has no attribute "Subscription"
samples/snippets/subscriber.py:290: error: Module "google.cloud.pubsub_v1.types" has no attribute "DeadLetterPolicy"
samples/snippets/subscriber.py:290: error: Module "google.cloud.pubsub_v1.types" has no attribute "FieldMask"
samples/snippets/subscriber.py:330: error: Module has no attribute "Subscription"
samples/snippets/subscriber.py:335: error: Redundant cast to "Subscription"
samples/snippets/subscriber.py:353: error: Module "google.cloud.pubsub_v1.types" has no attribute "FieldMask"
samples/snippets/subscriber.py:378: error: Module has no attribute "Subscription"
samples/snippets/subscriber.py:383: error: Redundant cast to "Subscription"
samples/snippets/subscriber.py:585: error: Call to untyped function "Retry" in typed context
samples/snippets/subscriber.py:631: error: Call to untyped function "Retry" in typed context
samples/snippets/publisher.py:281: error: Call to untyped function "Retry" in typed context
samples/snippets/publisher.py:286: error: Call to untyped function "if_exception_type" in typed context
samples/snippets/subscriber_test.py:179: error: Module "google.cloud.pubsub_v1.types" has no attribute "DeadLetterPolicy"
Found 17 errors in 4 files (checked 13 source files)
nox > Command mypy --config-file /tmpfs/src/github/python-pubsub/samples/snippets/mypy.ini --no-incremental samples/ failed with exit code 1
nox > Session mypy_samples failed.
```

owlbot.py
```diff
@@ -467,12 +412,12 @@ def mypy(session):
 # ----------------------------------------------------------------------------
 s.replace(
     "noxfile.py",
-    r'"pytype",',
+    r'    "mypy",',
     '\g<0>\n    # "mypy_samples",  # TODO: uncomment when the checks pass',
```
tswast, line 416: Out-of-date comment?

tswast (commented):
LGTM other than the failing samples checks. Thanks for adding the `api` property back.

owlbot.py
plamut: Indeed, will remove it so that owlbot will not add it back.

plamut:
> Getting a few mypy errors on the samples

This used to pass, will see what changes are needed to bring it back to green state.

plamut:
@tswast I confused the type-check errors with BigQuery, they were only fixed there.

There are two classes of errors reported by Pub/Sub samples type check:

- Generated proto types are injected into `google.cloud.pubsub_v1.types` dynamically at import time, and `mypy` cannot know about them.
  ```
  Module has no attribute "ProtoMessageType"
  ```
- Untyped objects from `python-api-core` - not all of them have been annotated, for example `api_core.retry.if_exception_type` and `api_core.retry.Retry` from the [retry module](https://github.com/googleapis/python-api-core/blob/74d5ce8a20aa6e270de75fa5ac37c0bb26df7b45/google/api_core/retry.py).
  ```
  Call to untyped function "if_exception_type" in typed context
  ```

While the latter can be solved by adding the missing annotations to `api-core`, the former would require quite some acrobatics to make it work.

We would have to avoid dynamically injected content, perhaps by some sort of a preprocessor that would copy/paste the generated types directly into `google/cloud/pubsub_v1/types.py`. In principle doable, but would require changes to the process and an update to tooling, which is too large a task to fit into this PR... :confused: 

cc: @pradn

tswast:
> While the latter can be solved by adding the missing annotations to api-core, the former would require quite some acrobatics to make it work.

I think we should definitely annotate the missing methods in `api-core` regardless of next steps.

> We would have to avoid dynamically injected content, perhaps by some sort of a preprocessor that would copy/paste the generated types directly into google/cloud/pubsub_v1/types.py

That is probably the right long-term solution. That said, it appears there are less than a dozen instances of "Module has no attribute" in the `mypy_samples` run. What if we use `google.pubsub_v1.types` instead of `google.cloud.pubsub_v1.types` in the code samples?

https://github.com/googleapis/python-pubsub/blob/main/google/pubsub_v1/types/__init__.py

The latter does have a few manual classes, but for the proto message types seems to just be redirecting (way too many in my opinion, there's no point in defining aliases for the shared types for example) various protobuf modules into the same namespace.

tswast:
Re: third option, since we haven't added a `py.typed` file to https://github.com/googleapis/python-pubsub/tree/main/google/cloud/pubsub_v1 yet, that'd be okay with me, as this is an improvement. We do need `mypy_samples` as a blocker for that, so let's make sure we cross-link https://github.com/googleapis/python-pubsub/issues/536 with the todo to uncomment `mypy_samples`.

tswast (approved):

PR #580, opened 2022-02-08
samples: create subscription with filtering enabled

anguillanneuf:
Towards b/214396829. 

The test output string looks something like `'Created subscription with filtering enabled: name: "projects/tz-playground-bigdata/subscriptions/subscription-test-su...: 604800\n}\nexpiration_policy {\n  ttl {\n    seconds: 2678400\n  }\n}\nfilter: "attributes.author=\\"unknown\\""\n\n'`, which is the reason for `assert '"attributes.author=\\"unknown\\""' in out` in the test.


hongalex (approved):

dandhlee (approved):
LGTM!

PR #585, opened 2022-02-14
fix(deps): move libcst to extras

busunkim96:
`libcst` is only needed to run the fixup script.


parthea (approved):
Should we also remove libcst here https://github.com/googleapis/python-pubsub/blob/main/testing/constraints-3.6.txt#L9?

busunkim96:
Extras are also included in the constraints file.

PR #600, opened 2022-03-03
fix(deps): require google-api-core>=1.31.5, >=2.3.2

parthea:
fix(deps): require proto-plus>=1.15.0


pradn:
@parthea What is the reason for this change?

pradn (approved):
I'd prefer a more descriptive PR comment, but it's fine if that's not possible.

parthea:
Thanks @pradn ! The relevant fix from google-api-core is https://github.com/googleapis/python-api-core/pull/318

PR #607, opened 2022-03-08
feat: retry temporary GRPC statuses for ack/modack/nack when exactly-once delivery is enabled

pradn:
We need to do this because [only UNAVAILABLE](https://github.com/googleapis/googleapis/blob/eb0700c6f29ca94f460307f201eb605744f055cb/google/pubsub/v1/pubsub_grpc_service_config.json#L221) is retried for acks/modacks/nacks at the GRPC level. With this CL, we extend the higher-level, manual retry mechanism for these RPCs to all the ones considered temporary for the Publish RPC. The new list of retriable codes is for these RPCs when exactly-once delivery is enabled is: DEADLINE_EXCEEDED, RESOURCE_EXHAUSTED, ABORTED, INTERNAL, UNAVAILABLE.

### DIFF
--- a/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
+++ b/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
@@ -75,13 +75,13 @@ _MIN_ACK_DEADLINE_SECS_WHEN_EXACTLY_ONCE_ENABLED = 60
 a subscription. We do this to reduce premature ack expiration.
 """
 
-_EXACTLY_ONCE_DELIVERY_TEMPORARY_RETRY_ERRORS = [
+_EXACTLY_ONCE_DELIVERY_TEMPORARY_RETRY_ERRORS = {
     code_pb2.DEADLINE_EXCEEDED,
     code_pb2.RESOURCE_EXHAUSTED,
     code_pb2.ABORTED,
     code_pb2.INTERNAL,
     code_pb2.UNAVAILABLE,
-]
+}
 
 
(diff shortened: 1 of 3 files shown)


noxfile.py
```diff
@@ -156,6 +156,7 @@ def default(session):
         "--cov-config=.coveragerc",
         "--cov-report=",
         "--cov-fail-under=0",
+        "-s",
```
acocuzzo, line 159: Assuming this will be removed before merge?

maheshgattani (approved):

acocuzzo (approved):

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -75,6 +75,14 @@
 a subscription. We do this to reduce premature ack expiration.
 """
 
+_EXACTLY_ONCE_DELIVERY_TEMPORARY_RETRY_ERRORS = [
```
mmicatka, line 78: Could we make this a set instead of an array?

noxfile.py
anguillanneuf: I believe this will be removed. `noxfile.py` is not to be edited.

anguillanneuf (changes requested):

pradn (commented):
Finished addressing Mike and Tianzi's comments.

anguillanneuf:
Lint errors will go away after you run `nox -s blacken`.

mmicatka (approved):
LGTM

anguillanneuf (approved):

PR #608, opened 2022-03-08
feat: return singleton success future for exactly-once methods in Message

pradn:
This is a small change to return a singleton success future that's constructed once instead of making a new future every time for *_with_response methods. We don't want to guarantee that we'll always return a distinct future object. This issue was raised in https://github.com/googleapis/python-pubsublite/pull/312.


dpcollins-google (approved):

mmicatka (approved):
LGTM

acocuzzo (approved):
LGTM

PR #619, opened 2022-03-18
samples(test): correct subscription type in exactly once delivery test

### LINKED ISSUE #616: Exactly once delivery sample not using a subscription with exactly once delivery enabled
https://github.com/googleapis/python-pubsub/blob/8f02850168fc9d7dee57f3b38f31d23d77c4aa27/samples/snippets/subscriber_test.py#L652-L657

anguillanneuf:
Fixes #616

### DIFF
--- a/samples/snippets/subscriber_test.py
+++ b/samples/snippets/subscriber_test.py
@@ -16,7 +16,7 @@ import os
 import re
 import sys
 import time
-from typing import Any, Callable, cast, Generator, TypeVar
+from typing import Any, Callable, cast, Generator, List, TypeVar
 import uuid
 
 from _pytest.capture import CaptureFixture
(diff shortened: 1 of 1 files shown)


anguillanneuf:
Test failure seems unrelated to the PR.

busunkim96 (approved):

samples/snippets/subscriber_test.py
```diff
@@ -307,7 +354,7 @@ def test_receive_with_delivery_attempts(
     # We keep retrying up to 10 minutes for mitigating the flakiness.
     @typed_backoff
     def run_sample() -> None:
-        _publish_messages(publisher_client, topic)
+        _ = _publish_messages(publisher_client, topic)
```
busunkim96, line 357: Does the return value need to be stored in `_`?

anguillanneuf: The returned messages IDs are only needed by one test. I changed them all because I added a return for the publish function.  `_` means in those other tests we don't care about the message IDs.

PR #626, opened 2022-03-28
fix: process ErrorInfo / GRPC errors for ack/modack only when exactly-once delivery is enabled

pradn:
* Don't process ErrorInfo / GRPC errors at all when exactly-once is disabled.
* Don't retry temporary errors for ack/modack RPCs when exactly-once is disabled
* Simplify completion of futures when a RetryError occurs (there's some duplication here but its simpler b/c we don't rely on subtle behavior in the process_errors function)
* Update tests / get to 100% test coverage


pradn:
Lint broken b/c https://github.com/psf/black/issues/2964

maheshgattani (approved):

acocuzzo (approved):
LGTM

PR #802, opened 2022-10-08
fix: batch at most 1,000 ack ids per request

beltran:
Brings [parity with the java client](https://github.com/googleapis/java-pubsub/blob/02aaa61635b37e9b822f701fba93d5e48706d81c/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java#L81). Allowing 2,500 ack ids in a single request can result in the request failing due to being too big.

Release-As: 2.13.10


acocuzzo:
Merging blocked till Oct 14 due to freeze.

acocuzzo (approved):
Will need to change expected calls in test_dispatcher splitting_large_payload acks and modacks tests from 3 to 6.

PR #881, opened 2023-03-10
samples: Schema evolution

kamalaboulhosn:


acocuzzo (approved):
Thanks so much, these look great! My only comments are regarding exceptions.
I may do a follow-up to add more exception types to new/existing samples.

samples/snippets/schema.py
```diff
@@ -359,6 +614,89 @@ def callback(message: pubsub_v1.subscriber.message.Message) -> None:
...
+        if encoding == "BINARY":
+            bout = io.BytesIO(message.data)
+            decoder = BinaryDecoder(bout)
+            message_data = reader.read(decoder)
+            print(f"Received a binary-encoded message:\n{message_data}")
+        elif encoding == "JSON":
+            message_data = json.loads(message.data)
+            print(f"Received a JSON-encoded message:\n{message_data}")
+        else:
+            print(f"Received a message with no encoding:\n{message}")
```
acocuzzo, line 681: should we nack here?

PR #1018, opened 2023-11-01
Suggested fix to exception due to Timeout argument being string type in snippet sample sub.py. 

### LINKED ISSUE #1017: Timeout argument throw exception as wrong type (str v.s. float expected) (Snippet sample)
Minor issue in "python-pubsub/samples/snippets/quickstart/sub.py"

#### Environment details

  - OS type and version:  macOS Sonoma 14.1
  - Python version: Python 3.10.11
  - pip version: pip 23.0.1
  - `google-cloud-pubsub` version: Version: 2.18.4

#### Steps to reproduce

  1. Start a simple pub/sub system on GCP (for example follow guide https://cloud.google.com/pubsub/docs/building-pubsub-messaging-system)
  2. as per the sample, instead of starting subscriber with "python sub.py $PROJECT sub_one" without a timeout argument, give it any timeout value. 
  3. subscriber will exit immediately due to exception, as timeout is in str type, un-matching float type expected in code "streaming_pull_future.result(timeout=timeout)"
  4. in addition, help message on line 60 is inaccurate.

#### Exception captured: 
'>' not supported between instances of 'str' and 'int'
```

### PR DESCRIPTION
**Pull request by changlichuan**
- [NA] Appropriate docs were updated (if necessary)

Fixes #1017 🦕

### DIFF
--- a/samples/snippets/quickstart/sub.py
+++ b/samples/snippets/quickstart/sub.py
@@ -42,7 +42,7 @@ def sub(project_id: str, subscription_id: str, timeout: Optional[float] = None) 
     try:
         # Calling result() on StreamingPullFuture keeps the main thread from
         # exiting while messages get processed in the callbacks.
-        streaming_pull_future.result(timeout=float(timeout))
+        streaming_pull_future.result(timeout=timeout and float(timeout))
     except:  # noqa
         streaming_pull_future.cancel()  # Trigger the shutdown.
         streaming_pull_future.result()  # Block until the shutdown is complete.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: samples/snippets/quickstart/sub.py [hunk 1]
```diff
@@ -42,7 +42,7 @@ def callback(message: pubsub_v1.subscriber.message.Message) -> None:
     try:
         # Calling result() on StreamingPullFuture keeps the main thread from
         # exiting while messages get processed in the callbacks.
-        streaming_pull_future.result(timeout=timeout)
+        streaming_pull_future.result(timeout=timeout and float(timeout))
```
pradn, line 45: Hi Changli, thanks for the bug and the fix. 

I think it'll be simpler if we remove the change to this line and instead add a "type=float" argument to the `add_argument` call below. `Argparse` already supports requiring arguments to be of a certain type. See [docs](https://docs.python.org/3/library/argparse.html#argparse-type).

pradn: I made a PR to keep this going forward. https://github.com/googleapis/python-pubsub/pull/1051

PR #1024, opened 2023-11-22
feat: Introduce compatibility with native namespace packages

parthea:
Similar to https://github.com/googleapis/google-auth-library-python/pull/1205
Also see https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages


parthea:
Adding `do not merge` for the release freeze. This should be merged after the release freeze ends on November 29th.

ohmayr (approved):

PR #1076, opened 2024-02-08
fix: add google-auth as a direct dependency

mukund-ananthu:
1. One of the issues this solves is the test coverage failures in https://github.com/googleapis/python-pubsub/actions/runs/7836080724/job/21382924522?pr=1064
2. pulling the google-auth dependency transitively through google-api-core so we decided to use the same `google-auth` version as `google-api-core` which is `2.14.1`
https://github.com/googleapis/python-api-core/blob/82c31184168d92a88ea570780082ab1176cfed53/setup.py#L34
3. Reference PR: https://github.com/googleapis/python-bigquery-storage/pull/740

- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-pubsub/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Fixes #<issue_number_goes_here> 🦕

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,6 +38,7 @@ else:
 dependencies = [
     "grpcio >= 1.51.3, < 2.0dev",  # https://github.com/googleapis/python-pubsub/issues/609
     # google-api-core >= 1.34.0 is allowed in order to support google-api-core 1.x
+    "google-auth >= 2.14.1, <3.0.0dev",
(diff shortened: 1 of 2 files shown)


hongalex (approved):

setup.py
michaelpri10, line 1: `< 3.0.dev0` to match https://github.com/googleapis/python-api-core/blob/82c31184168d92a88ea570780082ab1176cfed53/setup.py#L34

michaelpri10 (changes requested):

PR #1134, opened 2024-04-03
fix: Set timeout to infinite for publishing with ordering keys enabled

### LINKED ISSUE #1084: Publishing sometimes fails when ordering key is enabled
#### Environment details

- OS type and version: Debian Bookworm (Docker slim image)
- Python version: 3.9.18
- pip version: 23.1
- google-cloud-pubsub version: 2.18.4
- grpcio: 1.59.0

#### Description

Hello there,

We’ve encountered some publishing issues for a while now and ran out of ideas on how to fix them. The behaviour is the following. We publish messages into a few topics that are configured with an ordering key. Multiple times a week (or even a day), the publishing process fails and the client publisher stops (it seems to be stuck?), as no messages are published. On Google Cloud Monitoring, it usually states a ‘deadline exceeded’ and indeed, we do have timeouts when that happens (but no exception on the client side). This first observation is a bit weird though, given that we don’t publish massive loads of messages (just a few messages every second).

After (!) reading the [documentation](https://cloud.google.com/pubsub/docs/retry-requests?hl=fr#retry_ordering), this behaviour is supposed to be expected with ordering keys.

#### Code example

```python

# TODO(developer): Choose an existing topic.
# project_id = "your-project-id"
# topic_id = "your-topic-id"

publisher_options = pubsub_v1.types.PublisherOptions(
    enable_message_ordering=True,
    flow_control=PublishFlowControl(
        message_limit=2000, 
        limit_exceeded_behavior=LimitExceededBehavior.BLOCK
            ),)
# Sending messages to the same region ensures they are received in order
# even when multiple publishers are used.
client_options = {"api_endpoint": "us-east1-pubsub.googleapis.com:443"}
batch_settings = BatchSettings(
            1 * 1000 * 1000,
            0.01,
            1000,
        )
publisher = pubsub_v1.PublisherClient(
		batch_settings,
    publisher_options=publisher_options, 
		client_options=client_options
)
# The `topic_path` method creates a fully qualified identifier
# in the form `projects/{project_id}/topics/{topic_id}`
topic_path = publisher.topic_path(project_id, topic_id)

for message in [
    ("message1", "key1"),
    ("message2", "key2"),
    ("message3", "key1"),
    ("message4", "key2"),
]:
    # Data must be a bytestring
    data = message[0].encode("utf-8")
    ordering_key = message[1]
    # When you publish a message, the client returns a future.
    future = publisher.publish(topic_path, data=data, ordering_key=ordering_key)
    future.add_done_callback(resume_if_error)

def resume_if_error(future):
		try:
        print(future.result())
    except RuntimeError:
        # Resume publish on an ordering key that has had unrecoverable errors.
        publisher.resume_publish(topic_path, ordering_key)

print(f"Resumed publishing messages with ordering keys to {topic_path}.")
```
We wanted to avoid calling `result()` in the main thread as it is blocking, but use a callback instead. Anyway, this does not work, as we still have the same issues.

Can anyone help?

mukund-ananthu:
1. Setting timeout to be infinite when publishing with ordering keys enabled. 
2. This is required to retry infinitely when the client library receives retriable error codes from the backend.
3. Note that there is another test [test_publish_with_ordering_key_with_no_retry](https://github.com/googleapis/python-pubsub/blob/8c7e2a9e1b158e2b75dec8c5d78be17ce2dd449b/tests/unit/pubsub_v1/publisher/test_publisher_client.py#L320) that currently does not test the retry behavior or settings because the kwargs for `commit_retry` would be `None` and `commit_timeout` would be `_MethodDefault._DEFAULT_VALUE` when no retries are specified. This test will be improved in a subsequent PR.

Fixes #1084 🦕

### DIFF
--- a/google/cloud/pubsub_v1/publisher/client.py
+++ b/google/cloud/pubsub_v1/publisher/client.py
@@ -399,8 +399,13 @@ class Client(publisher_client.PublisherClient):
                     transport = self._transport
                     base_retry = transport._wrapped_methods[transport.publish]._retry
                     retry = base_retry.with_deadline(2.0**32)
+                    # timeout needs to be overridden and set to infinite in
+                    # addition to the retry deadline since both determine
+                    # the duration for which retries are attempted.
+                    timeout = 2.0**32
                 elif retry is not None:
                     retry = retry.with_deadline(2.0**32)
+                    timeout = 2.0**32
 
             # Delegate the publishing to the sequencer.
             sequencer = self._get_or_create_sequencer(topic, ordering_key)
--- a/tests/unit/pubsub_v1/publisher/test_publisher_client.py
+++ b/tests/unit/pubsub_v1/publisher/test_publisher_client.py
@@ -316,6 +316,10 @@ def test_publish_with_ordering_key_uses_extended_retry_deadline(creds):
     expected_retry = custom_retry.with_deadline(2.0**32)
     _assert_retries_equal(batch_commit_retry, expected_retry)
 
+    batch_commit_timeout = kwargs["commit_timeout"]
+    expected_timeout = 2.0**32
+    assert batch_commit_timeout == pytest.approx(expected_timeout)
+
 
 def test_publish_with_ordering_key_with_no_retry(creds):
     client = publisher.Client(


google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -399,8 +399,10 @@ def on_publish_done(future):
                     transport = self._transport
                     base_retry = transport._wrapped_methods[transport.publish]._retry
                     retry = base_retry.with_deadline(2.0**32)
+                    timeout = 2.0**32
```
pradn, line 402: Add comment to explain what parameter this is. Add comment to explain why we set this to infinite (not obvious).

pradn (approved):

PR #1185, opened 2024-06-06
chore: Update mypy version

### LINKED ISSUE #1061: Upgrade to the latest version of mypy
The current version of mypy in `noxfile.py` is `0.910` which is more than 2 years old. We should update to the latest version.
https://github.com/googleapis/python-pubsub/blob/265f4106f499ec5d2d01a127ba192404c1836a28/noxfile.py#L35C1-L35C1

mukund-ananthu:
Current version is `0.910` which is more than 2 years old. Update mypy version to the latest available version in PyPi. Additionally fix all the type checking errors currently detected and reported by mypy.

Fixes #1061 🦕

### DIFF
--- a/google/cloud/pubsub_v1/subscriber/scheduler.py
+++ b/google/cloud/pubsub_v1/subscriber/scheduler.py
@@ -162,7 +162,7 @@ class ThreadScheduler(Scheduler):
                 work_item = self._executor._work_queue.get(block=False)
                 if work_item is None:  # Exceutor in shutdown mode.
                     continue
-                dropped_messages.append(work_item.args[0])
+                dropped_messages.append(work_item.args[0])  # type: ignore[index]
         except queue.Empty:
             pass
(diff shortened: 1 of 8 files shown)


parthea (approved):

google/cloud/pubsub_v1/subscriber/scheduler.py
```diff
@@ -162,7 +162,7 @@ def shutdown(
                 work_item = self._executor._work_queue.get(block=False)
                 if work_item is None:  # Exceutor in shutdown mode.
                     continue
-                dropped_messages.append(work_item.args[0])
+                dropped_messages.append(work_item.args[0])  # type: ignore[index]
```
parthea, line 165: Can you add a comment to provide more detail on why we are ignoring the type error? If there is follow up work, please create a new issue and link to it here.

PR #1193, opened 2024-06-14
chore: Remove restriction that protobuf should be less than 5

### LINKED ISSUE #1192: Unable to use patched version of grpcio with google-cloud-pubsub
There are a number of bugs (functionally and security wise) in grpcio < 1.63. However, >1.63 now requires >5 protobuf, but this library requires protobuf <5 making it no longer possible to install this library with a patched working version of grpcio.

```
Because no versions of google-cloud-pubsub match >2.21.3
 and google-cloud-pubsub (2.21.3) depends on protobuf (>=3.19.5,<3.20.0 || >3.20.0,<3.20.1 || >3.20.1,<4.21.0 || >4.21.0,<4.21.1 || >4.21.1,<4.21.2 || >4.21.2,<4.21.3 || >4.21.3,<4.21.4 || >4.21.4,<4.21.5 || >4.21.5,<5.0.0dev), google-cloud-pubsub (>=2.21.3) requires protobuf (>=3.19.5,<3.20.0 || >3.20.0,<3.20.1 || >3.20.1,<4.21.0 || >4.21.0,<4.21.1 || >4.21.1,<4.21.2 || >4.21.2,<4.21.3 || >4.21.3,<4.21.4 || >4.21.4,<4.21.5 || >4.21.5,<5.0.0dev).
Because grpcio (1.64.1) depends on protobuf (>=5.26.1,<6.0dev)
 and no versions of grpcio match >1.64.1,<2.0.0, grpcio (>=1.64.1,<2.0.0) requires protobuf (>=5.26.1,<6.0dev).
Thus, google-cloud-pubsub (>=2.21.3) is incompatible with grpcio (>=1.64.1,<2.0.0).
```

Can support for a newer version of protobuf be included in this library?

mukund-ananthu:
I see that an upper bound of 5 was set on the protobuf library in https://github.com/googleapis/python-pubsub/pull/762/commits/23ce35e92c68f561fd2208439479983f0f8344c8 

I do not find any documentation on reasons for keeping an upper bound. Proposing removing the upper bound in this PR, but also checking if there are any reasons we would not want to do this from @parthea who would have historical context on this.

Fixes #1192 🦕

BEGIN_COMMIT_OVERRIDE
fix: allow Protobuf 5.x
END_COMMIT_OVERRIDE


jeffsawatzky:
I would very much like to have this fixed as well, for similar reasons to #1192
@mukund-ananthu / @parthea this is currently blocking me due to dependency hell. If you could remove this restriction (or loosen it to <6.0.0dev for now) it would make things much easier. Thanks!

setup.py
```diff
@@ -42,7 +42,7 @@
     "google-api-core[grpc] >= 1.34.0, <3.0.0dev,!=2.0.*,!=2.1.*,!=2.2.*,!=2.3.*,!=2.4.*,!=2.5.*,!=2.6.*,!=2.7.*,!=2.8.*,!=2.9.*,!=2.10.*",
     "proto-plus >= 1.22.0, <2.0.0dev",
     "proto-plus >= 1.22.2, <2.0.0dev; python_version>='3.11'",
-    "protobuf>=3.19.5,<5.0.0dev,!=3.20.0,!=3.20.1,!=4.21.0,!=4.21.1,!=4.21.2,!=4.21.3,!=4.21.4,!=4.21.5",
+    "protobuf>=3.19.5,!=3.20.0,!=3.20.1,!=4.21.0,!=4.21.1,!=4.21.2,!=4.21.3,!=4.21.4,!=4.21.5",
```
parthea, line 45: Please add an upper bound `<6.0.0` so that users aren't broken when a new major version of protobuf is released. Also , bump the minimum version to `protobuf>=3.20.2` which matches what we have in `python-api-common-protos`

https://github.com/googleapis/python-api-common-protos/blob/c77c0dc5d29ef780d781a3c5d757736a9ed09674/setup.py#L26-L28

```suggestion
    "protobuf>=3.20.2,<6.0.0dev,!=4.21.0,!=4.21.1,!=4.21.2,!=4.21.3,!=4.21.4,!=4.21.5",
```

Otherwise LGTM

parthea (approved):

parthea:
Also update `testing/constraints-3.7.txt` 

Change

https://github.com/googleapis/python-pubsub/blob/0b8d7b91da7928e8f06379d19c616e7a6a411349/testing/constraints-3.7.txt#L10

to 

```
protobuf==3.20.2
```

parthea:
The system tests are failing due to an unrelated issue. 
See https://github.com/googleapis/python-pubsub/issues/created_by/app/flaky-bot

```
___________ test_subscriber_not_leaking_open_sockets[rest-rest-rest] ___________

publisher = 
topic_path_base = 'projects/<REDACTED>/topics/<REDACTED>'
subscription_path_base = 'projects/<REDACTED>/subscriptions/<REDACTED>'
cleanup = [(>, (), {'topic': 'projects/<REDACTED>/topics/<REDACTED>'})]
transport = 'rest'

    @pytest.mark.parametrize("transport", ["grpc", "rest"])
    def test_subscriber_not_leaking_open_sockets(
        publisher, topic_path_base, subscription_path_base, cleanup, transport
    ):
        # Make sure the topic and the supscription get deleted.
        # NOTE: Since subscriber client will be closed in the test, we should not
        # use the shared `subscriber` fixture, but instead construct a new client
        # in this test.
        # Also, since the client will get closed, we need another subscriber client
        # to clean up the subscription. We also need to make sure that auxiliary
        # subscriber releases the sockets, too.
        custom_str = "-not-leaking-open-sockets"
        subscription_path = subscription_path_base + custom_str
        topic_path = topic_path_base + custom_str
        subscriber = pubsub_v1.SubscriberClient(transport="grpc")
        subscriber_2 = pubsub_v1.SubscriberClient(transport="grpc")
    
        cleanup.append(
            (subscriber_2.delete_subscription, (), {"subscription": subscription_path})
        )
        cleanup.append((subscriber_2.close, (), {}))
        cleanup.append((publisher.delete_topic, (), {"topic": topic_path}))
    
        # Create topic before starting to track connection count (any sockets opened
        # by the publisher client are not counted by this test).
        publisher.create_topic(name=topic_path)
    
        current_process = psutil.Process()
>       conn_count_start = len(current_process.connections())

tests/system.py:471: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = psutil.Process(pid=397, name='py.test', status='running', started='16:34:59')
args = (), kwargs = {}

    @functools.wraps(fun)
    def inner(self, *args, **kwargs):
>       warnings.warn(msg, category=DeprecationWarning, stacklevel=2)
E       DeprecationWarning: connections() is deprecated and will be removed; use net_connections() instead
```

PR #1252, opened 2024-09-16
feat: Add OpenTelemetry support for Subscribe Side

mukund-ananthu:
Subscribe Side OpenTelemetry Code

### DIFF
--- a/google/cloud/pubsub_v1/open_telemetry/subscribe_opentelemetry.py
+++ b/google/cloud/pubsub_v1/open_telemetry/subscribe_opentelemetry.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from typing import Optional
+from datetime import datetime
 
 from opentelemetry import trace
 from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
@@ -67,3 +68,12 @@ class SubscribeOpenTelemetry:
             end_on_exit=False,
         ) as subscribe_span:
             self._subscribe_span = subscribe_span
+
+    def add_subscribe_span_event(self, event: str) -> None:
+        assert self._subscribe_span is not None
+        self._subscribe_span.add_event(
+            name=event,
+            attributes={
+                "timestamp": str(datetime.now()),
+            },
+        )
--- a/google/cloud/pubsub_v1/open_telemetry/subscribe_opentelemetry.py
+++ b/google/cloud/pubsub_v1/open_telemetry/subscribe_opentelemetry.py
@@ -159,3 +159,7 @@ class SubscribeOpenTelemetry:
             end_on_exit=False,
         ) as process_span:
             self._process_span = process_span
+
+    def end_process_span(self) -> None:
+        assert self._process_span is not None
+        self._process_span.end()
--- a/google/cloud/pubsub_v1/open_telemetry/subscribe_opentelemetry.py
+++ b/google/cloud/pubsub_v1/open_telemetry/subscribe_opentelemetry.py
@@ -1,0 +1,69 @@
+# Copyright 2024, Google LLC All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Optional
+
+from opentelemetry import trace
+from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
+
+from google.cloud.pubsub_v1.open_telemetry.context_propagation import (
+    OpenTelemetryContextGetter,
+)
+from google.pubsub_v1.types import PubsubMessage
+
+
+class SubscribeOpenTelemetry:
+    _OPEN_TELEMETRY_TRACER_NAME: str = "google.cloud.pubsub_v1"
+    _OPEN_TELEMETRY_MESSAGING_SYSTEM: str = "gcp_pubsub"
+
+    def __init__(self, message: PubsubMessage):
+        self._message: PubsubMessage = message
+
+        # subscribe span will be initialized by the `start_subscribe_span`
+        # method.
+        self._subscribe_span: Optional[trace.Span] = None
+
+    def start_subscribe_span(
+        self,
+        subscription: str,
+        exactly_once_enabled: bool,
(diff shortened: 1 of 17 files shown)


google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -252,6 +268,10 @@ def ack(self) -> None:
             https://cloud.google.com/pubsub/docs/exactly-once-delivery."
 
         """
+        if self.opentelemetry_data:
+            self.opentelemetry_data.add_subscribe_span_event("ack start")
+            self.opentelemetry_data.add_process_span_event("ack called")
```
hongalex, line 273: This span should be added when the user calls msg.Ack(), rather than when the internal library ack is called.

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
```diff
@@ -303,6 +316,10 @@ def drop(
         Args:
             items: The items to drop.
         """
+        for item in items:
+            if item.opentelemetry_data:
+                item.opentelemetry_data.set_subscribe_span_result("dropped")
```
hongalex, line 321: I think we call this "expired" in the design, but if you prefer dropped to match the client library method, that's fine.

google/cloud/pubsub_v1/subscriber/message.py
hongalex: Ok reading this, it's the subscribe span event that is actually in the wrong place. That should start with the RPC is invoked, not when the user calls msg.Ack.

google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -429,12 +467,17 @@ def nack(self) -> None:
         may take place immediately or after a delay, and may arrive at this subscriber
         or another.
         """
+        if self.opentelemetry_data:
+            self.opentelemetry_data.add_subscribe_span_event("nack start")
```
hongalex, line 471: Same comment as ack, event should be added when the RPC is invoked

google/cloud/pubsub_v1/subscriber/message.py
```diff
@@ -406,6 +439,8 @@ def modify_ack_deadline_with_response(self, seconds: int) -> "futures.Future":
             will be thrown.
 
         """
+        if self.opentelemetry_data:
+            self.opentelemetry_data.add_subscribe_span_event("modack start")
```
hongalex, line 443: Same as above: make this event when the RPC is called.

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -1110,6 +1150,7 @@ def _on_response(self, response: gapic_types.StreamingPullResponse) -> None:
                 )
                 return
 
+            i: int = 0
             for received_message in received_messages:
```
hongalex, line 1154: Not a python expert, but does using enumerate work better here instead of incrementing the counter manually?

google/cloud/pubsub_v1/subscriber/message.py
mukund-ananthu: Got it. Moved the `add_subscribe_span_event("ack start")` from message.py to dispatcher.py(where the actual RPC call is made). Did this for both ack and ack_with response. Similarly for nack / nack_with_response and modify_ack_deadline and modify_ack_deadline_with_response

google/cloud/pubsub_v1/subscriber/_protocol/dispatcher.py
mukund-ananthu: Sure, renamed to "expired"

google/cloud/pubsub_v1/subscriber/message.py
mukund-ananthu: Done. Response https://github.com/googleapis/python-pubsub/pull/1252#discussion_r1769706317

google/cloud/pubsub_v1/subscriber/message.py
mukund-ananthu: Done. Response: https://github.com/googleapis/python-pubsub/pull/1252#discussion_r1769706317

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
mukund-ananthu: Reason for not trying to enumerate here is that `subscribe_opentelemetry[]` can either be having the same length as the received_messages list when opentelemetry is enabled, in which case an enumeration can be used, but `subscribe_opentelemetry` is an empty list when opentelemetry is disabled. In this case, enumeration would not work, because received_messages and subscribe_opentelemetry would have N and zero length, and something along the lines of:
`for x, y in zip(received_messages, subscribe_opentelemetry)` would cause issues. Hence, I used just a normal counter here.

mukund-ananthu:
@parthea PTAL

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
```diff
@@ -1007,22 +1020,84 @@ def _get_initial_request(
...
+        warn_on_invalid=True,
+        receipt_modack: bool = False,
     ) -> Set[str]:
         exactly_once_enabled = False
+
+        modack_span: Optional[trace.Span] = None
+        if self._client.open_telemetry_enabled:
+            subscribe_span_links: List[trace.Link] = []
+            subscribe_spans: List[trace.Span] = []
+            assert len(self._subscription.split("/")) == 4
```
hongalex, line 1036: nit: split this just once and save the result

google/cloud/pubsub_v1/open_telemetry/subscribe_opentelemetry.py
```diff
@@ -0,0 +1,279 @@
...
+        ack_id: str,
+        delivery_attempt: int,
+    ) -> None:
+        tracer = trace.get_tracer(_OPEN_TELEMETRY_TRACER_NAME)
+        parent_span_context = TraceContextTextMapPropagator().extract(
+            carrier=self._message,
+            getter=OpenTelemetryContextGetter(),
+        )
+        self._publisher_create_span_context = parent_span_context
+        assert len(subscription.split("/")) == 4
```
hongalex, line 88: nit: same as above, save split result

hongalex (approved):

google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py
mukund-ananthu: Done.

google/cloud/pubsub_v1/open_telemetry/subscribe_opentelemetry.py
mukund-ananthu: Done

PR #1258, opened 2024-09-25
feat: Add OpenTelemetry publish sample

mukund-ananthu:
Fixes #<issue_number_goes_here> 🦕

### DIFF
--- a/samples/snippets/publisher.py
+++ b/samples/snippets/publisher.py
@@ -170,6 +170,83 @@ def delete_topic(project_id: str, topic_id: str) -> None:
     # [END pubsub_delete_topic]
 
 
+def pubsub_publish_otel_tracing(
+    topic_project_id: str, trace_project_id: str, topic_id: str
+) -> None:
+    """
+    Publish to `topic_id` in `topic_project_id` with OpenTelemetry enabled.
+    Export the OpenTelemetry traces to Google Cloud Trace in project
+    `trace_project_id`
+
+    Args:
+        topic_project_id: project ID of the topic to publish to.
+        trace_project_id: project ID to export Cloud Trace to.
+        topic_id: topic ID to publish to.
+
+    Returns:
+        None
+    """
+    # [START pubsub_publish_otel_tracing]
+
+    from opentelemetry import trace
+    from opentelemetry.sdk.trace import TracerProvider
+    from opentelemetry.sdk.trace.export import (
+        BatchSpanProcessor,
+    )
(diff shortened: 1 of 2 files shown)


samples/snippets/publisher.py
```diff
@@ -170,6 +170,87 @@ def delete_topic(project_id: str, topic_id: str) -> None:
...
+    # topic_project_id = "your-topic-project-id"
+    # trace_project_id = "your-trace-project-id"
+    # topic_id = "your-topic-id"
+
+    # In this sample, we use a Google Cloud Trace to export the OpenTelemetry
+    # traces: https://cloud.google.com/trace/docs/setup/python-ot
+    # Choose and configure the exporter for your set up accordingly.
+
+    # Sample 1 in every 1000 traces.
+    sampler = ParentBased(root=TraceIdRatioBased(1 / 1000))
```
hongalex, line 212: Can you sample at 100% to begin with and add a comment to lower the trace in production?

I can imagine customers who use the sample don't see that there are any traces exported and think something is broken.

samples/snippets/publisher.py
```diff
@@ -170,6 +170,87 @@ def delete_topic(project_id: str, topic_id: str) -> None:
...
+            enable_open_telemetry_tracing=True,
+        ),
+    )
+
+    # The `topic_path` method creates a fully qualified identifier
+    # in the form `projects/{project_id}/topics/{topic_id}`
+    topic_path = publisher.topic_path(topic_project_id, topic_id)
+
+    tracer = trace.get_tracer("google.cloud.pubsub_v1.publisher")
+    with tracer.start_as_current_span("parent cloud trace span"):
```
hongalex, line 239: I think we should omit this, since it's not necessary to wrap the traces in this way.

samples/snippets/publisher.py
mukund-ananthu: I use 1/100 with 10k messages published. That ensures there would be atleast 100 samples exported.

samples/snippets/publisher.py
mukund-ananthu: Done. This would mean, there would be no parent trace. But as discussed offline, we are good with this / this is intended.

samples/snippets/publisher.py
mukund-ananthu: Done, changed it to 1 based on offline discussion.

hongalex (approved):

PR #1273, opened 2024-10-09
docs: Add ingestion from GCS sample

michaelpri10:
Adds samples for creating an ingestion from GCS topic.


samples/snippets/publisher.py
```diff
@@ -103,6 +103,74 @@ def create_topic_with_kinesis_ingestion(
...
+    from google.cloud import pubsub_v1
+    from google.protobuf import timestamp_pb2
+    from google.pubsub_v1.types import Topic
+    from google.pubsub_v1.types import IngestionDataSourceSettings
+
+    # TODO(developer)
+    # project_id = "your-project-id"
+    # topic_id = "your-topic-id"
+    # bucket = "your-bucket"
+    # input_format = "text"
```
mukund-ananthu, line 126: Is text the only input_format that is supported, or should we make this explicit to the user that they could set the input_format based on their needs.

michaelpri10: I've added a comment for the possible input_format types.

mukund-ananthu (approved):

PR #1326, opened 2025-01-11
fix: Stop using api_core default timeouts in publish since they are broken

### LINKED ISSUE #1325: Python api_core has a broken timeout logic. Put in an interim fix before they get to fixing it.
Refer https://github.com/googleapis/python-api-core/issues/654

mukund-ananthu:
### Change in behavior

Default publish timeout behavior:

Pre-change: 60s timeout for initial RPC, 0s timeout per retry, 600s total timeout
Post-change:60s timeout for initial RPC, 60s timeout per retry, 600s total timeout

### Change

Only non-test / non sample(i.e functional) usage found for `PublisherOptions.timeout` is here:

https://github.com/googleapis/python-pubsub/blob/f648f65624d6096908a1e9c4364b36aaf928ce11/google/cloud/pubsub_v1/publisher/client.py#L440-L441

This change in usage is intended.

### Call stack:

https://github.com/googleapis/python-pubsub/blob/f648f65624d6096908a1e9c4364b36aaf928ce11/google/cloud/pubsub_v1/publisher/client.py#L299

invokes one of the sequencers(ordered or unordered)

https://github.com/googleapis/python-pubsub/blob/f648f65624d6096908a1e9c4364b36aaf928ce11/google/cloud/pubsub_v1/publisher/client.py#L477-L479

which then create a batch with this timeout

https://github.com/googleapis/python-pubsub/blob/f648f65624d6096908a1e9c4364b36aaf928ce11/google/cloud/pubsub_v1/publisher/_sequencer/ordered_sequencer.py#L255-L263

which uses it when committing the batch / publishing the messages:

https://github.com/googleapis/python-pubsub/blob/f648f65624d6096908a1e9c4364b36aaf928ce11/google/cloud/pubsub_v1/publisher/_batch/thread.py#L323-L328

and invokes gapic_publish: https://github.com/googleapis/python-pubsub/blob/f648f65624d6096908a1e9c4364b36aaf928ce11/google/cloud/pubsub_v1/publisher/client.py#L289-L291

which then uses it in the RPC call:
https://github.com/googleapis/python-pubsub/blob/f648f65624d6096908a1e9c4364b36aaf928ce11/google/pubsub_v1/services/publisher/client.py#L1067-L1072

that is obtained by wrapping the transport base publish call:

https://github.com/googleapis/python-pubsub/blob/f648f65624d6096908a1e9c4364b36aaf928ce11/google/pubsub_v1/services/publisher/client.py#L1055

The wrapper uses the timeout when invoking the GapicCallable:

https://github.com/googleapis/python-api-core/blob/b1fae31c8c71ed65ad22a415dab2b54720c3d4ba/google/api_core/gapic_v1/method.py#L245-L252

which currently overrides float and int timeouts to use TimeToDeadlineTimeout:

https://github.com/googleapis/python-api-core/blob/b1fae31c8c71ed65ad22a415dab2b54720c3d4ba/google/api_core/gapic_v1/method.py#L112-L113

which is incorrect(why its incorrect is mentioned in the attached bug)

Fixes #1325 🦕

### DIFF
--- a/google/cloud/pubsub_v1/types.py
+++ b/google/cloud/pubsub_v1/types.py
@@ -35,6 +35,7 @@ from google.protobuf import field_mask_pb2
 from google.protobuf import timestamp_pb2
 
 from google.api_core.protobuf_helpers import get_messages
+from google.api_core.timeout import ConstantTimeout
 
 from google.pubsub_v1.types import pubsub as pubsub_gapic_types
 
@@ -191,7 +192,10 @@ class PublisherOptions(NamedTuple):
         "an instance of :class:`google.api_core.retry.Retry`."
     )
 
-    timeout: "OptionalTimeout" = gapic_v1.method.DEFAULT  # use api_core default
+    # Use ConstantTimeout instead of api_core default because the default
+    # value results in retries with zero deadline.
+    # Refer https://github.com/googleapis/python-api-core/issues/654
+    timeout: "OptionalTimeout" = ConstantTimeout(60)
     (
         "Timeout settings for message publishing by the client. It should be "
         "compatible with :class:`~.pubsub_v1.types.TimeoutType`."
--- a/tests/unit/pubsub_v1/publisher/test_publisher_client.py
+++ b/tests/unit/pubsub_v1/publisher/test_publisher_client.py
@@ -19,6 +19,7 @@ import inspect
 import sys
 
 import grpc
+import math
 
 # special case python < 3.8
 if sys.version_info.major == 3 and sys.version_info.minor < 8:
@@ -35,6 +36,7 @@ from opentelemetry import trace
 from google.api_core import gapic_v1
 from google.api_core import retry as retries
 from google.api_core.gapic_v1.client_info import METRICS_METADATA_KEY
+from google.api_core.timeout import ConstantTimeout
 
 from google.cloud.pubsub_v1 import publisher
 from google.cloud.pubsub_v1 import types
@@ -652,6 +654,8 @@ def test_publish_new_batch_needed(creds):
     future = client.publish(topic, b"foo", bar=b"baz")
     assert future is mock.sentinel.future
 
+    call_args = batch_class.call_args
+
     # Check the mocks.
     batch_class.assert_called_once_with(
         client=mock.ANY,
@@ -660,8 +664,12 @@ def test_publish_new_batch_needed(creds):
         batch_done_callback=None,
         commit_when_full=True,
         commit_retry=gapic_v1.method.DEFAULT,
-        commit_timeout=gapic_v1.method.DEFAULT,
+        commit_timeout=mock.ANY,
     )
+    commit_timeout_arg = call_args[1]["commit_timeout"]
+    assert isinstance(commit_timeout_arg, ConstantTimeout)
+    assert math.isclose(commit_timeout_arg._timeout, 60) is True
+
     message_pb = gapic_types.PubsubMessage(data=b"foo", attributes={"bar": "baz"})
     wrapper = PublishMessageWrapper(message=message_pb)
     batch1.publish.assert_called_once_with(wrapper)


mukund-ananthu:
beltran@ PTAL

hongalex (approved):

google/cloud/pubsub_v1/types.py
```diff
@@ -191,7 +192,12 @@ class PublisherOptions(NamedTuple):
         "an instance of :class:`google.api_core.retry.Retry`."
     )
 
-    timeout: "OptionalTimeout" = gapic_v1.method.DEFAULT  # use api_core default
+    # Use ExponentialTimeout instead of api_core default because the default
+    # value results in retries with zero deadline.
+    # Refer https://github.com/googleapis/python-api-core/issues/654
+    timeout: "OptionalTimeout" = ExponentialTimeout(
+        initial=5, maximum=60, multiplier=1.3, deadline=600
```
kamalaboulhosn, line 199: I'm a little nervous about changing the default behavior from 60s to 5s for the initial request. Prior to the change that resulted in unexpected backoff behavior, what was the initial value? If it was 5, this could be considered a return to the "correct" value, but if this is the first time we are introducing it as 5, I fear we could break behavior for people.

mukund-ananthu: >Prior to the change that resulted in unexpected backoff behavior, what was the initial value?

Initial value = 60s


### Pre api_core change:

initial RPC value = 60s
retry RPC value = 60s
total timeout across retries = 600s

### Post api_core change:

initial RPC value = 60s
retry RPC value = 0s
total timeout across retries = 600s

### Post this PR:

Initial RPC value = 5s
retry RPC value = [5, 60] with multiplier=1.3


>but if this is the first time we are introducing it as 5, I fear we could break behavior for people.

* This is the first time we have 5s as the initial RPC value and yes, this is a breaking change.
* Hypothetically, even if the initial value was 5s(it isn't), then [changed](https://github.com/googleapis/python-api-core/pull/462) to 60s in 2022, I'm not sure how much weight we must give to the fact that it was 5s initially, since its been a few years with this new behavior, i.e, this would still be a breaking change for those who onboarded between 2022 and now. 


### Code / Links for easy reference

* Both pre and post api_core [change](https://github.com/googleapis/python-api-core/pull/462), we [passed in](https://github.com/plamut/python-pubsub/blob/23e1612b402f3f9dae247f5f0f4e6756d14f08b0/google/pubsub_v1/services/publisher/transports/base.py#L151) `default_timeout=60` from Pub/Sub to api_core library.
* Pre api_core change, this was interpreted as [ConstantTimeout(60)](https://github.com/googleapis/python-api-core/blob/b1fae31c8c71ed65ad22a415dab2b54720c3d4ba/google/api_core/timeout.py#L159)
* Post api_core change, this is interpreted as  [TimeToDeadlineTimeout(60)](https://github.com/googleapis/python-api-core/blob/b1fae31c8c71ed65ad22a415dab2b54720c3d4ba/google/api_core/timeout.py#L118)
* This interpretation of `default_timeout=60.0` in api_core happens [here](https://github.com/googleapis/python-api-core/blob/b1fae31c8c71ed65ad22a415dab2b54720c3d4ba/google/api_core/gapic_v1/method.py#L112-L113)
* Post PR change, timeout decorator applied would be [ExponentialTimeout](https://github.com/googleapis/python-api-core/blob/b1fae31c8c71ed65ad22a415dab2b54720c3d4ba/google/api_core/gapic_v1/method.py#L112-L116) instead of [TimeToDeadlineTimeout(60)](https://github.com/googleapis/python-api-core/blob/b1fae31c8c71ed65ad22a415dab2b54720c3d4ba/google/api_core/timeout.py#L118) 
* Definition of different timeout options can be found [here](https://github.com/googleapis/python-api-core/blob/b1fae31c8c71ed65ad22a415dab2b54720c3d4ba/google/api_core/timeout.py#L1)

kamalaboulhosn (approved):

PR #1348, opened 2025-01-23
fix: set creds only if transport not provided

ohmayr:
GAPICs are designed to allow users to only configure one of transport or credentials. If a transport is configured explicitly, then it must be instantiated with credentials.

gapic-generator: See: https://github.com/googleapis/gapic-generator-python/blob/main/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/client.py.j2#L563.

pubsub: https://github.com/googleapis/python-pubsub/blob/main/google/pubsub_v1/services/publisher/client.py#L678

in google.cloud.pubsub_v1, if the environment variable `PUBSUB_EMULATOR_HOST` is set, the credentials are automatically set to `AnonymousCredentials` in the [client layer](https://github.com/googleapis/python-pubsub/blob/main/google/cloud/pubsub_v1/publisher/client.py#L135).

This means that if `PUBSUB_EMULATOR_HOST` is set, then credentials are automatically configured under the hood for test cases.

This behaviour results in an issue if a test sets `PUBSUB_EMULATOR_HOST` explicitly and provides a transport to a client, resulting in the following error to be raised https://github.com/googleapis/python-pubsub/blob/main/google/pubsub_v1/services/publisher/client.py#L678.

Given that we want to update the following [test](https://github.com/googleapis/python-pubsub/blob/main/tests/unit/pubsub_v1/publisher/test_publisher_client.py#L417) to explicitly configure a transport to the client (with an intercept channel), we want to make sure that credentials are provided to the transport instance and not on the client instance directly.


google/cloud/pubsub_v1/publisher/client.py
```diff
@@ -132,7 +132,9 @@ def __init__(
             kwargs["client_options"] = {
                 "api_endpoint": os.environ.get("PUBSUB_EMULATOR_HOST")
             }
-            kwargs["credentials"] = AnonymousCredentials()
+            # Configure credentials directly to transport, if provided.
```
parthea, line 135: Add a todo for https://github.com/googleapis/python-pubsub/issues/1349 here as well

google/pubsub_v1/services/publisher/client.py
```diff
@@ -713,6 +713,8 @@ def __init__(
             )
             # initialize with the provided callable or the passed in class
 
+            # TODO(https://github.com/googleapis/python-pubsub/issues/1349): Move the emulator
+            # code below to test files.
```
parthea, line 717: I believe this file is autogenerated. We can modify the file by changing the code below

https://github.com/googleapis/python-pubsub/blob/8adb133a8caa0744980ffc393331fe4fdda2e3b4/owlbot.py#L102-L116

parthea (approved):

PR #1513, opened 2025-10-22
fix: Check span recording state before ending process span

juandelacruz-calvo:
## Summary

This PR adds a check to verify that the subscribe span is still recording before attempting to end the process span in the `__exit__` method of the `SubscribeOpenTelemetry` context manager.

## Problem

When the OpenTelemetry subscriber context manager exits, it may attempt to end the process span even when the parent subscribe span has already been terminated or is no longer in a recording state. This leads to log errors like:

[/.venv/lib/python3.12/site-packages/opentelemetry/sdk/trace/__init__.py#L943](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/3327b42f786af8c37b49469f154d4c3e10ac1770/.venv/lib/python3.12/site-packages/opentelemetry/sdk/trace/__init__.py#L943)
```py
                logger.warning("Calling end() on an ended span.")
```

This happens because using `pubsub` instrumentation the `end()` method may be called twice:

- On `ack` https://github.com/googleapis/python-pubsub/blob/e6294a1883abf9809cb56d5cd4ad25cc501bc994/google/cloud/pubsub_v1/subscriber/message.py#L276

- Then on `__exit__` https://github.com/googleapis/python-pubsub/blob/e6294a1883abf9809cb56d5cd4ad25cc501bc994/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py#L174-L175

## Solution

Added a condition `self._subscribe_span.is_recording()` to the existing check before calling `self.end_process_span()`:

```python
if self._process_span and self._subscribe_span.is_recording():
    self.end_process_span()
```

## Testing

- Existing unit tests pass
- Integration tests with OpenTelemetry context propagation pass

## Impact

This is a defensive code change that prevents edge case errors in span management. It should not affect normal operation but provides better handling when spans are terminated in non-standard flows.


abbrowne126 (approved):

abbrowne126:
Thanks for putting this together!

